PR #7, opened 2015-01-08
Add ability to request current app.

SarenCurrie:
Creates a publishable event to get the currently loaded app. This returns said app to a callback in the form
```
{
  id: "123456789",
  name: "example"
}
```


JRMaitre:
Good work and thank you! 

Out of curiosity, what are you using this code for? :)

PR #14, opened 2015-02-22
Dutch Language

yumminova:
Complete, after i get Monster working (i cold need a hand with that) i will test, some things are imposible to translate without a running version ))


JRMaitre:
Hello, Thank you for your pull-request! However, if you check the files changed, it appears you're submitting non-valid JSON. You need to escape the " symbol in some places. I recommend running your files in a JSON Validator like http://jsonlint.com/ before submitting your pull-request. It also appears that you submitted a '.directory' file, and I don't know what it is but it doesn't belong in the project! Please fix those changes and submit a new pull-request! :)

PR #17, opened 2015-03-01
Show API error message in dialog popup from kazoo SDK.

Bat-o-matic:
When using the Kazoo SDK to make API requests in Monster (e.g. via self.callApi), the dialog popup on request error only shows a message related to the error code (e.g. "Access forbidden." for 403).

This PR displays the API error message in the dialog popup as well.


JRMaitre:
Thank you for the PR. We're going to add an error log in Monster soon, you can check ticket UI-1074 to see what's coming if you have any remark!

PR #22, opened 2015-09-09
Fixed issue with error messages in monster.request

SarenCurrie:
Error data wasn't being proliferated through when error wasn't a 402, parsedError variable wasn't being assigned in other branches.


JRMaitre:
Thank you, good catch.

PR #24, opened 2015-11-14
Fix mixed/blocked content browser warning when GUI is served via https

bhepp:


JRMaitre:
Thanks, we missed that :)

PR #26, opened 2015-11-21
UI-1918 use reseller's own service plans

bhepp:
Adds additional argument 'useOwnPlans', which is set to true for resellers. This allows a reseller to assign his own service plans instead of his parent's service plans. Affects only the "Add New Account" wizard.

See the corresponding pull request for the monster-ui-accounts repo (https://github.com/2600hz/monster-ui-accounts/pull/8)


JRMaitre:
Thank you for the fix :)

PR #27, opened 2016-01-27
fix(accounts): Coerce numerics to strings during search

rdpfeffer:
Its possible that the data stored in the DOM, and retrieved using jQuery's data() function could return a numeric value. By casting this to a string before calling toLowerCase() we prevent any exception that would break search. This is possible in cases where the data returned is of the form...

```json
      {
          "id": "asdfasfdsadsasdf",
          "name": 1234,
          "realm": "1234.something.com"
      }
```

### DIFF
--- a/apps/common/submodules/accountBrowser/accountBrowser.js
+++ b/apps/common/submodules/accountBrowser/accountBrowser.js
@@ -147,7 +147,7 @@ define(function(require){
 				var found = false;
 
 				_.each(data, function(val) {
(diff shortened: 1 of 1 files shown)


rdpfeffer:
Thanks for adding in the ability to masquerade via deep links. This https://github.com/2600hz/monster-ui/commit/59639352fdbda4fa37651a29aef48bd502a30053 will definitely come in handy for us!

JRMaitre:
Thank you!

And glad you enjoy the masquerading stuff, it's pretty ghetto, but pretty nice for support & such :)

PR #48, opened 2016-11-29
Numbers formatting US Regex should only match US numbers

mariovokal:
Hi,
this fix should prevent strange formatting of non US numbers, with length between 10 and 12 digits.
For example, without this, number "+3531xxxyyzz", will be turned into: "+3+1 (531) xxx-yyzz"
Thanks,
Mario


JRMaitre:
Thank you!

PR #64, opened 2016-12-15
Ensure jquery is loaded before everything that requires it

mira-t:


azefiel:
Thank you for your contribution!

PR #67, opened 2017-01-15
Update app.js

powerpbx:
Add case of logging in with ApiKey generated tokens.

### DIFF
--- a/src/apps/auth/app.js
+++ b/src/apps/auth/app.js
@@ -105,10 +105,12 @@ define(function(require){
 			}
 			// Otherwise, we check if some GET parameters are defined, and if they're formatted properly
 			
-			//APIkey generated tokens also require UserId parameter to login.
+			//APIkey generated tokens require UserId parameter to login.
 			else if(urlParams.hasOwnProperty('u') && urlParams.hasOwnProperty('t')) {
-				self.authenticateGET(urlParams, successfulAuth, errorAuth);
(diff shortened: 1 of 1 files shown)


JRMaitre:
Hello! Thanks for the PR, and it's a good point, we need a user id.

After checking the code, I think it would make more sense to use the same authenticateAuthToken method, but to change the success callback by something like:

```
/* Stuff */
else if(urlParams.hasOwnProperty('u') && urlParams.hasOwnProperty('t')) {
    self.authenticateAuthToken(urlParams.t, function(authData) {
        authData.data.owner_id = urlParams['u'];
        successFulAuth(authData);
    }, errorAuth);
}
// Username/password generated tokens do not require anything else to log in.
else if(urlParams.hasOwnProperty('t')) {
/* Stuff */
```

instead of referencing successfulAuth directly and using another GET request.

Does that make sense?

powerpbx:
Glad you are open to adding this feature.  Your way makes more sense.

JRMaitre:
Do you want to update the commit with the fix and I'll merge it down? :)

powerpbx:
Not sure is your urlParams['u'] or urlParams.u is considered more correct.  Both seem to work so I went with the latter to make it consistent with `urlParams.t`

JRMaitre:
Yep you're right! Thank you for the PR :)

PR #80, opened 2017-03-06
UI-2570: add steps and friendly printer to slider control

JRMaitre:


azefiel:
Also, not related to the content of this PR but, a slider seems overkill to select between five values... A dropdown would be more appropriate imo.

JRMaitre:
Thanks a lot for looking at it. Good feedback :)

As for the "slider" vs "select", I think it's actually nice to show sliders when we can and it's appropriate as they look nice and are a little more user-friendly than dropdowns. I didn't like it at first, but I like it there :)

PR #94, opened 2017-04-26
Do not try to parse empty string as JSON in response error.

bpar476:


JRMaitre:
Thanks for the PR! :)

PR #115, opened 2017-07-29
more images

plau:
adds images missing from this branch that is in master


azefiel:
Hey @plau, could you explain why certain images both have a jpg and png version?

For example, `vtech_vsp6xx_vsp600` and `vtech`.

PR #126, opened 2017-09-26
Update portListing.js

arekfryz:
dateObj.getMonth() is an integer number, between 0 and 11. You should add 1 in order to get correct date string.


azefiel:
Thank you for contributing @arekfryz

PR #128, opened 2017-09-26
Bring port app dialog to the front

arekfryz:
Quick fix to bring port app dialog to the front. Currently is covered by modal with default z-index:1000; Maybe it could be attached to different element in order to fix that too. This is quick fix.


azefiel:
Thanks a bunch for spotting that issue.

We came up with a more generic solution (97c13d069b8084730047b36bb747d12b2f6b5316) and therefor will close this PR.

PR #130, opened 2017-09-28
Fix race condition when binding multiple listeners to websocket

mira-t:
When two bindings are created in close succession (to the same routing key), one can overwrite the other.


JRMaitre:
Thanks that's a good catch :)

PR #134, opened 2017-09-30
underscorejs use _.extend instead of _.merge

arekfryz:
I see you replaced lodash with underscorejs _.merge needs to be replaced with _.extend


azefiel:
Thanks for your contribution @arekfryz.

PR #143, opened 2017-10-13
UI-2932: Remove nicescroll library

azefiel:


azefiel:
Corrupted main.js with missing core libraries.

PR #145, opened 2017-10-16
Replace _.merge with _.extend

arekfryz:


azefiel:
Hi,

Thank you for reporting this issue @arekfryz, it was caused by a commit (a86cdeb1) that do not belong on this branch.

I will revert this commit to correct the problem.

PR #157, opened 2018-01-18
Add alternative apploader

mira-t:
This change would allow an alternative apploader to be used, if desired, by enabling and configuring it through settings in config.js. This alternative apploader is a drop-down menu of apps, which is faster to select from and could be a more practical option for accounts with few apps. Additionally, it is possible to choose whether or not to expose the App Store in the drop-down - if not enabled via config.js, it will be hidden from admins too.

### DIFF
--- a/src/apps/apploader/app.js
+++ b/src/apps/apploader/app.js
@@ -59,7 +59,7 @@ define(function(require) {
 			var self = this;
 
 			if (monster.config.whitelabel.useDropdownApploader) {
-				return $('#apploader').length !== 0;
+				return $('.app-list-dropdown-wrapper').length !== 0;
 			} else {
 				return typeof self.appFlags.modal !== 'undefined';
 			}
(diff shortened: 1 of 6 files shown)


JRMaitre:
Hi and thank you for the PR!

On the first look it looks like a good idea. I'll have to test it a little more before merging it.

I noticed you pushed your config file, and you should probably remove it from the PR :)

Cheers,
JR

mira-t:
Thank you for pointing that out! I have removed the part of the config that shouldn't be added. I left the other part inside because it indicates how to enable the alternative apploader.

JRMaitre:
Thank you! I'll review the rest tomorrow :)

src/apps/apploader/app.js
```diff
@@ -71,38 +75,69 @@ define(function(require) {
...
-					var template = $(monster.template(self, 'app', {
-						allowAppstore: monster.apps.auth.currentUser.priv_level === 'admin',
-						defaultApp: monster.ui.formatIconApp(appList[0]),
-						apps: appList
-					}));
+					if (monster.config.whitelabel.useDropdownApploader) {
+						template = $(monster.template(self, 'appList', {
+							defaultApp: appList[0],
+							apps: appList,
+							useDropdownApploader: true,
```
JRMaitre: I don't think this is needed, useDropdownApploader is unused in that template

src/apps/apploader/app.js
```diff
@@ -71,38 +75,69 @@ define(function(require) {
...
-						hideClose: true,
-						destroyOnClose: false
-					});
+					if (!monster.config.whitelabel.useDropdownApploader) {
+						self.appFlags.modal = monster.ui.fullScreenModal(template, {
+							hideClose: true,
+							destroyOnClose: false
+						});
+					}
+
```
JRMaitre: Please remove blank line

JRMaitre:
Hi,

After I started to review I ended up coding a few changes to follow our ESLint guidelines and also cleanup some code that didn't need to be there. Can you let me know if these changes work for you? I tested locally and it works fine but just want to make sure before I go ahead and merge your PR with my updates :)

Thanks,
JR

JRMaitre:
I was also wondering why you need the showAppstoreInDropdown? I wonder if we could remove this from this PR as it's more of a flag that should work across both "types" of apploader if we were to add it.

mira-t:
Thank you very much for taking the time to review and improve my pull request!

I tested your changes with different combinations of the two new flags in the config file and different users (with and without admin privileges) and the behaviour is as expected. i.e. useDropdownApploader being set to true always shows the dropdown apploader for all users and showAppstoreInDropdown being set to true makes the appstore link appear if the dropdown apploader is used and the user is an admin.

I added the flag showAppstoreInDropdown because in our usage of Monster UI at Conversant, every user already has all the apps they have access to and we do not want to expose the app store, even if the user is an admin. I didn't make this flag apply to both types of apploader because of scope but you are right, it should apply to both types of apploader if we were to add it. Would you like me to remove it and create a separate pull request for hiding the appstore link?

JRMaitre:
Yes that would be great :)

JRMaitre:
Thank you for the PR 👍

PR #159, opened 2018-01-23
Add preflow selection to account settings

mira-t:
This would allow users to select an existing callflow as an account-level preflow through Monster UI.

### DIFF
--- a/src/apps/myaccount/app.js
+++ b/src/apps/myaccount/app.js
@@ -940,6 +940,10 @@ define(function(require) {
 						params.data.ui_flags.numbers_format_exceptions = newData.ui_flags.numbers_format_exceptions;
 					}
 				}
+				// Remove preflow when user sets it to be none
+				if(params.data.preflow && params.data.preflow.always && params.data.preflow.always === 'none') {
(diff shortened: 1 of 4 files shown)


JRMaitre:
Hi!

Did you know you could set this setting in the Callflows app? (Under account settings > misc > preflow settings)

If yes, is it just because you thought it'd be faster and easier to expose it in the MyAccount section?

The code looks good to me, but I'm not sure that this is a setting we want to expose there, so I'll have to ask our product team to see if it's OK. As a general rule of thumb we tend to only expose settings that most people use in the MyAccount section.

Thanks for the PR, and I'll get back to you soon!
JR

mira-t:
Apologies, I did not know this was now available through the callflows app. Thank you very much for your help!

JRMaitre:
No worries! Feel free to ask us questions via the community or IRC before developing it yourself, I hate to waste your work! :(

mira-t:
I originally made this as a customization for a previous version of monster that didn't expose preflows, so I already had the changes and only needed to move it in and make sure it was compatible :) Next time I will have a better check if things have already been added!

PR #161, opened 2018-01-25
Access lists

Nathan-Cairns:
Added section for access lists in myAccount app. 
A user can now change their access lists from there.


JRMaitre:
Hi and thank you for the contribution.

Since it's a lot of code, can you please make sure you follow our linting standards? I noticed that it doesn't. If you want instructions you can review it here: https://docs.2600hz.com/ui/docs/linting/

Let me know if you need help setting it up or have any question.

Cheers,
JR

Nathan-Cairns:
Hi JR,

Sweet as. I'll do that now.

Cheers,
Nathan

JRMaitre:
Hey Nathan,

Kind of a good news-bad news for you. The bad news is that we don't really see this as part of the My Account app, as we're trying to keep the settings simple and choose the ones that apply to most users. The good news is that we're very excited to see this code get into Monster as we haven't had the chance yet to build anything around access_lists.

After talking to our product team, we think the best fit for this feature would be to be its own application. The same way we have a Webhooks app or a Dialplan manager app, we could have an Access list app! 

What are your thoughts on moving this to its own app? You could create your own git repo following our "application" structure and build it there, or we could create a git repo for you and accept your PRs on it. Let me know if you have any questions or concerns about this process :)

Cheers,
JR

Nathan-Cairns:
Hi JR,

Good to hear you like the feature!

However, I'm concerned that it won't be enough content to constitute its own app.
It's just a drop down and a textbox and this could look a bit empty.

What are your thoughts on making it configurable in config.js?
This way it could be an optional addition to customers who need it.
It would also keep the settings menu nice and simple for those who don't need it!

Cheers,

Nathan

JRMaitre:
I think there's potential to expose more than a dropdown, and you could expose the device settings as well in an app of its own, as well as adding a help/tutorial section on how to use it. That being said, I understand that it's a lot more work to add to your feature.

If you want to leave it as is, then you can add a boolean flag in config.js in the whitelabel object to make it show up when set to true. If the flag is missing or set to any other value, it should hide this section.

I'll do another review after that 👍 

Thank you!

Nathan-Cairns:
Sweet as!
That should be done now.
Let me know if you've got any issues with it.

Thanks!

JRMaitre:
One thing I forgot to mention, is that since the APIs you're using are from Kazoo, you should add them to the src/js/lib/jquery.kazoosdk.js following the naming standards we use with all APIs (you can create an object "accessList" with the get/post/update API for account and devices). You would then remove the monster.requests from your code, and use self.callApi instead. You can then empty the "requests" object at the top of your app. Monster.request is intended to be used only for APIs external to Kazoo (so an API call to an analytics service, a google map API, etc...). Hope this makes sense, let me know if you have any questions :)

Nathan-Cairns:
Sweet as @JRMaitre that should be all good! Let me know if there's anything else.
Cheers.

JRMaitre:
Thank you for the contribution!

zachbwh:
Wow awesome job Nathan!

PR #164, opened 2018-02-11
Hide/show new account button via monster config

mira-t:
It is not always desirable to expose account creation to users. This commit adds the option of hiding it via a new option in the whitelabel section of the monster config.


BorigTheDwarf:
Any updates on this?

JRMaitre:
Sorry about the delay. It works well and I think it makes sense to add it as it's a feature we've been asked for before and is also opt-in only. I'll merge!

Thank you for the contribution as always 👍

kalda341:
@JRMaitre What's the status of this?

PR #165, opened 2018-02-20
Add setting to prevent sending of credentials to new user

mira-t:
Accompanying pull request for [https://github.com/2600hz/monster-ui-voip/pull/47](https://github.com/2600hz/monster-ui-voip/pull/47)


JRMaitre:
Hi! If people don't want to send the credentials to new user, then they have to uncheck the box in the User Creation dialog, this isn't something that we want to expose in the config file for now. Sorry about that!

Cheers,

PR #220, opened 2018-10-25
UI-3184: Display owner type for numbers in use across all view types

guillegr123:


azefiel:
This needs a more in depth product review than I thought.

Closing the PR for now and keeping the feature branch alive for future reference.

guillegr123:
Side to side view.

image

PR #222, opened 2018-10-29
UI-3155: Do not display talking blue shadow when user is muted

guillegr123:


src/apps/common/submodules/conferenceViewer/conferenceViewer.js
```diff
@@ -374,6 +375,9 @@ define(function(require) {
 
 					break;
 				case 'start-talking':
+					if ($userDiv.find('[data-action-category="mute"]').hasClass('active')) {
```
azefiel: Here you can reverse the test:
```suggestion
					if (!$userDiv.find('[data-action-category="mute"]').hasClass('active')) {
```

src/apps/common/submodules/conferenceViewer/conferenceViewer.js
```diff
@@ -374,6 +375,9 @@ define(function(require) {
...
+						break;
```
azefiel: and replace your additional `break` statement by adding the class instead:
```suggestion
						$userDiv.addClass('speaking');
```

src/apps/common/submodules/conferenceViewer/conferenceViewer.js
```diff
@@ -374,6 +375,9 @@ define(function(require) {
...
+					}
 					$userDiv.addClass('speaking');
```
azefiel: and finally get ride of this statement.
```suggestion

```

PR #224, opened 2018-11-01
UI-3215: Refactor `randomString()` using lodash to enforce type check and throw errors

joristirado:


src/js/lib/monster.util.js
```diff
@@ -1245,6 +1234,48 @@ define(function(require) {
 			);
 	}
 
+	/**
+	 * Generates a string of `length` random characters chosen from either a
+	 * preset or a custom string of characters.
+	 * @param  {Number} length  Number of characters to include.
+	 * @param  {String} pPreset Characters to choose from.
```
guillegr123: Javascript primitive types start with lowercase: https://github.com/airbnb/javascript#types
Also, is there no problem if this parameter name differs from the one that is specified in the MD document?

src/js/lib/monster.util.js
```diff
@@ -1245,6 +1234,48 @@ define(function(require) {
...
+	 * @return {String}         A string of random characters.
+	 */
+	function randomString(length, pPreset) {
+		if (!_.isNumber(length)) {
+			throw new TypeError('Expected a number for `length`');
+		}
+		if (!_.isUndefined(pPreset) && !_.isString(pPreset)) {
```
guillegr123: What should happen if `pPreset` is an empty string? Currently `monster.util.randomString(5, '')` returns an empty string.

src/js/lib/monster.util.js
azefiel: The section you are referring to details how primitive values are stored into variables (deep instead of shallow copy), I don't see a reference to lower casing.

Arguments that start with a lower case `p` usually mean they are optional, useful for type checking and allow us to declare a variable with the actual name:
```javascript
function getLength(array, pMultiplier) {
    // checks on mandatory `array` argument
    // ...

    // check on optional argument
    var multiplier = _.isUndefined(pMultiplier)
        ? 1
        : pMultiplier;

    // now you can use `multiplier` in any cases
}
```
We usually use this technique to assign default values to optional arguments.

Since it does not really matter for developers to know that's how optional arguments are handled internally,  in the doc we remove the leading `p` from optional arguments and simply display the name.

There is other ways to specifically declare optional arguments in the doc:

- in the function syntax section: anything between `[]` is optional
- in the parameters section: the `Required` column

P.S.: Please try to refer to [our style guide](https://docs.2600hz.com/ui/docs/javascriptStyleGuide/) instead of Airbnb's as it contains modifications to match Monster UI's codebase.

src/js/lib/monster.util.js
azefiel: Good catch!

In my opinion, that's the expected behavior of the method, as it does what it's told to do, namely, pick five characters out of an empty set of characters.

One improvement from this would be to check if the `preset` is empty and return an empty string right away instead of starting the chaining operation.

src/js/lib/monster.util.js
```diff
@@ -1245,6 +1234,48 @@ define(function(require) {
...
+			numerals: '1234567890',
+			hex: '1234567890abcdef',
+			safe: '23456789abcdefghjkmnpqrstuvwxyz'
+		};
+		var preset = _
+			.chain(presets)
+			.get(string, string)
+			.shuffle()
+			.value();
+		var getRandomItem = function() {
```
guillegr123: There could be a single `var` for all the variables declaration (a catch from @masmerino13)

src/js/lib/monster.util.js
```diff
@@ -1245,6 +1234,48 @@ define(function(require) {
...
+			var isUpper = _.sample([true, false]);
+			var item = preset[_.random(preset.length - 1)];
```
guillegr123: Here too could be a single variable declaration.

PR #233, opened 2018-11-11
Fix incorrect reference to documented method in description

teemak:
Changed lined 27 to monster.getTemplate() instead of monster.template().


azefiel:
Thanks for spotting that!

As I was scanning through the doc, I realized the description for `ignoreSpaces` in the parameters table is incorrect and should read:
```
When set to `true`, carriage return, linefeed, tab, whitespace will not be trimmed from the template.
```
Do you think you could update that too as part of this PR?

teemak:
Done.

azefiel:
Please add your changes from #235 here, we don't need multiple PRs to fix simple typos.

PR #235, opened 2018-11-12
TYPO

teemak:
Removed a slash before whitelabel anchor.


azefiel:
Closed in favor of #233

PR #243, opened 2018-11-22
UI-3252: Contextually load port requests for account/descendants

azefiel:


masmerino13:
[FIX](https://github.com/2600hz/monster-ui/pull/245)

PR #247, opened 2018-11-26
UI-3257: Allow super-duper user to post private comments at port request

guillegr123:
portingcomment-01
portingcomment-02
portingcomment-03

### DIFF
--- a/src/apps/common/submodules/portListing/portListing.js
+++ b/src/apps/common/submodules/portListing/portListing.js
@@ -780,7 +780,8 @@ define(function(require) {
 						formattedComment = {
 							timestamp: moment(monster.util.gregorianToDate(newComment.timestamp)).valueOf(),
 							title: newComment.author,
-							content: newComment.content
+							content: newComment.content,
+							isSuperDuperEntry: newComment.superduper_comment
 						};
 
 					// check if the last entry was created the same day than today
@@ -820,7 +821,8 @@ define(function(require) {
 				timeline.push({
 					timestamp: moment(monster.util.gregorianToDate(comment.timestamp)).valueOf(),
 					title: comment.author,
-					content: comment.content
+					content: comment.content,
+					isSuperDuperEntry: comment.superduper_comment
 				});
 			});
 
@@ -1089,6 +1091,10 @@ define(function(require) {
 			event.preventDefault();
 
 			var comment = $textarea.val();
+
+			if (_.chain(comment).trim().isEmpty().value()) {
+				return;
+			}
 
 			self.portListingHelperAddComment(_.merge({}, args, {
 				data: {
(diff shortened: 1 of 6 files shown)


src/apps/common/submodules/portListing/portListing.js
```diff
@@ -552,21 +554,21 @@ define(function(require) {
...
 
-			template
+			textareaWrapper
 				.find('#add_comment')
 					.on('click', function(event) {
-						event.preventDefault();
-
-						var textarea = template.find('textarea[name="comment"]'),
-							comment = textarea.val();
+						self.portListingHelperPostComment(event, textarea, args, false);
```
masmerino13: Looks good,

Please consider to send parameters like an object, also I think you don't need to send the `onClick event` to `portListingHelperPostComment`, also with that way should be pretty simple tu define a default value for `isSuperDuperComment`
```suggestion
						event.preventDefault();

						self.portListingHelperPostComment({
							textarea: textarea,
							args: args
						});
```

src/apps/common/submodules/portListing/portListing.js
```diff
@@ -552,21 +554,21 @@ define(function(require) {
...
+					});
 
-						self.portListingHelperAddComment($.extend(true, {}, args, {
-							data: {
-								comment: comment
-							}
-						}));
+			if (!monster.util.isSuperDuper()) {
+				this.return;
```
masmerino13: Is this ok?
```suggestion
return;
```

src/apps/common/submodules/portListing/portListing.js
```diff
@@ -552,21 +554,21 @@ define(function(require) {
...
+			}
 
-						textarea.val('');
+			// Set event handlers for superduper options
+			textareaWrapper
+				.find('#add_private_comment')
+					.on('click', function(event) {
+						self.portListingHelperPostComment(event, textarea, args, true);
```
masmerino13: Sending parameters like an object, this should looks like this:
```suggestion
						event.preventDefault();

						self.portListingHelperPostComment({
							textarea: textarea,
							args: args,
							isSuperDuperComment: true
						});
```

src/apps/common/submodules/portListing/portListing.js
```diff
@@ -745,6 +747,27 @@ define(function(require) {
 			}
 		},
 
+		portListingHelperPostComment: function(event, $textarea, args, isSuperDuperComment) {
+			var self = this;
```
masmerino13: And here:
```suggestion
			var self = this,
				$textarea = params.textarea,
				args = params.args,
				isSuperDuperComment = _.get(params, 'isSuperDuperComment', false);
```

PR #251, opened 2018-12-03
UI-3220: Don't hide gear icon on spare numbers list when delete or move a number

masmerino13:


masmerino13:
@azefiel fixed on `move to` action also

src/apps/common/submodules/numbers/numbers.js
```diff
@@ -1072,8 +1058,11 @@ define(function(require) {
 			});
 		},
 
-		numbersPaintSpare: function(parent, dataNumbers, callback) {
+		numbersPaintSpare: function(args) {
```
guillegr123: This function could be renamed to `numbersRenderSpare`, to follow the naming practices on other submodules for functions that display parts of the view.

src/apps/common/submodules/numbers/numbers.js
```diff
@@ -69,13 +64,13 @@ define(function(require) {
 
 				var arrayNumbers = data.listAccounts.length ? data.listAccounts[0].usedNumbers : [];
```
guillegr123: Maybe a function similar to `numbersPaintSpare` could be created for used numbers, to reduce repeating code.

src/apps/common/submodules/numbers/numbers.js
```diff
@@ -1072,8 +1058,11 @@ define(function(require) {
 			});
 		},
 
-		numbersPaintSpare: function(parent, dataNumbers, callback) {
+		numbersPaintSpare: function(args) {
 			var self = this,
+				parent = args.parent,
```
guillegr123: This variable can be removed, as `parent` will be used only once within the function.
```suggestion
```

src/apps/common/submodules/numbers/numbers.js
```diff
@@ -1085,6 +1074,9 @@ define(function(require) {
 				.empty()
 				.append(template);
 
+			var arrayNumbersSpare = dataNumbers.listAccounts.length ? dataNumbers.listAccounts[0].spareNumbers : [];
+			self.numbersDisplayFeaturesMenu(arrayNumbersSpare, template);
+
 			callback && callback();
```
guillegr123: To follow previous practices, this could be changed:
```suggestion
			args.hasOwnProperty('callback') && args.callback();
```

src/apps/common/submodules/numbers/numbers.js
masmerino13: `used numbers` list is needed once on the first load

src/apps/common/submodules/numbers/numbers.js
masmerino13: yep, sounds good!

src/apps/common/submodules/numbers/numbers.js
guillegr123: You are right, it is true

guillegr123: I forgot this one, the `arrayNumbers` declaration can be moved to the previous variable declaration block (https://docs.2600hz.com/ui/docs/javascriptStyleGuide/#variables).

src/apps/common/submodules/numbers/numbers.js
```diff
@@ -1072,20 +1058,24 @@ define(function(require) {
...
 				}));
 
-			parent
+			args.parent
 				.find('.list-numbers[data-type="spare"]')
 				.empty()
 				.append(template);
 
-			callback && callback();
+			var arrayNumbersSpare = dataNumbers.listAccounts.length ? dataNumbers.listAccounts[0].spareNumbers : [];
```
guillegr123: The variable `arrayNumbersSpare` should be declared and assigned at the top of the scope, along with the other variable declarations, according to https://docs.2600hz.com/ui/docs/javascriptStyleGuide/#variables

vccs:
It looks like this has resurfaced when deleting has an error...

image

PR #252, opened 2018-12-04
UI-3261: Close popup and refresh row after caller ID change

guillegr123:


src/apps/common/submodules/callerId/callerId.js
```diff
@@ -28,15 +41,24 @@ define(function(require) {
...
 			var self = this,
+				dataNumber = args.numberData,
 				popup_html = $(self.getTemplate({
 					name: 'layout',
 					data: dataNumber.cnam || {},
 					submodule: 'callerId'
 				})),
 				popup,
-				accountId = pAccountId || self.accountId,
+				accountId = args.accountId || self.accountId,
```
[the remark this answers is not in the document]
guillegr123: Applied :smiley:

src/apps/common/submodules/callerId/callerId.js
```diff
@@ -12,11 +12,24 @@ define(function(require) {
...
+		 * @param  {Function} [args.callbacks.success]  Success callback
+		 * @param  {Function} [args.callbacks.error]    Error callback
+		 */
 		callerIdEdit: function(args) {
 			var self = this,
 				argsCommon = {
-					success: function(dataNumber) {
-						self.callerIdRender(dataNumber, args.accountId, args.callbacks);
+					success: function(numberData) {
+						self.callerIdRender(_.assign({
```
azefiel: Using `lodash#assign` here can be misleading since you just want to add `success`/`error` to this object I believe, Unless you want to override `numberData`?
```suggestion
						self.callerIdRender(_.merge({
```

src/apps/common/submodules/callerId/callerId.js
```diff
@@ -51,36 +73,75 @@ define(function(require) {
...
 				ev.preventDefault();
 
-				if (monster.ui.valid(form)) {
-					var cnamFormData = monster.ui.getFormData('cnam');
+				if (!monster.ui.valid(form)) {
+					return;
+				}
 
-					_.extend(dataNumber, { cnam: cnamFormData });
+				_.extend(dataNumber, {
```
azefiel: `lodash#extend` is legacy code from the migration to lodash (from underscore).

Although it [achieves the intended goal](https://lodash.com/docs/#assignIn) in this case, let's try to move away from it locally and use the more appropriate method (only assign own props vs own + inherited props).
```suggestion
				_.assign(dataNumber, {
```

src/apps/common/submodules/callerId/callerId.js
```diff
@@ -51,36 +73,75 @@ define(function(require) {
...
-								})
-							});
-
-							popup.dialog('destroy').remove();
-
-							callbacks.success && callbacks.success(data);
-						},
-						function(data) {
-							callbacks.error && callbacks.error(data);
+				monster.waterfall([
```
azefiel: Nice clean up!

src/apps/common/submodules/callerId/callerId.js
```diff
@@ -93,28 +154,62 @@ define(function(require) {
...
-					data: data
+				data: args.data,
+				success: function(data) {
+					args.hasOwnProperty('success') && args.success(data);
 				},
-				success: function(_data, status) {
-					if (typeof success === 'function') {
-						success(_data);
+				error: function(parsedError) {
+					if (parsedError.error === '402') {
```
azefiel: Now that I see this, I realize it is not ideal to have to guard against `402` errors for potentially each request handler.

Maybe we should handle this in the SDK directly so that the `error` callback is not called on such HTTP errors, I'll create a ticket for this!

src/apps/common/submodules/callerId/callerId.js
guillegr123: Thanks :smile:

src/apps/common/submodules/callerId/callerId.js
guillegr123: Sounds good, maybe with an option similar to the `generateError` flag? (to allow the charges event to be handled by the SDK or the error callback).

PR #269, opened 2019-01-24
UI-3296: Include Telicon v2 icons, and create helpers to render included SVG icons

guillegr123:


src/js/lib/monster.ui.js
```diff
@@ -275,6 +275,15 @@ define(function(require) {
 				return monster.ui.getTemplatePhoneNumber(number.toString());
 			},
 
+			renderSvgIcon: function(id, cssClass) {
+				var template = monster.ui.getTemplateSvgIcon({
+					id: id,
+					cssClass: _.isString(cssClass) ? cssClass : undefined
+				});
+
+				return new Handlebars.SafeString(template);
```
masmerino13: If this is just a wrapper of `getTemplateSvgIcon` I think this line should be moved to `monster.ui.getTemplateSvgIcon` and just return his result:
```
return monster.ui.getTemplateSvgIcon({
  id: id,
  cssClass: _.isString(cssClass) ? cssClass : undefined
});
```

let me know if i'm wrong :)

src/js/lib/monster.ui.js
```diff
@@ -3093,6 +3102,36 @@ define(function(require) {
...
+			throw TypeError('"args.cssClass" is not a string');
+		}
+		var iconId = args.id;
+		var iconPrefix = iconId.substring(0, iconId.indexOf('--'));
+		var templateData = {
+			iconId: iconId,
+			cssClass: _.get(args, 'cssClass', 'svg-icon ' + iconPrefix)
+		};
+
+		return monster.template(monster.apps.core, 'monster-svg-icon', templateData);
```
masmerino13: To sence this to return a handlebars template: 
`return new Handlebars.SafeString(monster.template(monster.apps.core, 'monster-svg-icon', templateData));`

let me know if i'm wrong :)

src/css/lib/svg-icons/svg-icons.scss
```diff
@@ -0,0 +1,29 @@
+.svg-icon {
+	&.g-drive {
+		/* Default size for icons 16px = 1em */
+		width: 6.429em;	// Scaled based on main symbol's viewBox size: 200×1.125em÷35
```
masmerino13: `rem` vs `em` :thinking:

src/js/lib/monster.ui.js
```diff
@@ -3093,6 +3102,36 @@ define(function(require) {
...
 
+	/**
+	 * Get handlebars templateto render an SVG icon
+	 * @param   {Object} args
+	 * @param   {String} args.id                   Icon ID
+	 * @param   {String} [args.cssClass=svg-icon]  CSS classes to be applied to the SVG tag
+	 * @return  {String}                           SVG icon template
+	 */
+	function getTemplateSvgIcon(args) {
+		if (!_.isPlainObject(args)) {
```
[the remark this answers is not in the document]
guillegr123: `else` is not necessary because the function finishes with the `throw` statement, and also to follow the convention used in the new functions like `randomString` at monster.util.js.

src/js/lib/monster.ui.js
guillegr123: I was considering it, but looking at the code in monster.ui.js, only the handlebars helpers create `Handlebars.SafeString` objects, so I think it can be kept this way.

src/js/lib/monster.ui.js
guillegr123: This function, on the other hand, is not a Handlebars helper, so it may be OK to return just the template as string, like `getTemplatePhoneNumber`.

src/css/lib/svg-icons/svg-icons.scss
guillegr123: Not sure. I set `em` to allow to resize the icon according to its container, instead of globally, like [FontAwesome](https://github.com/FortAwesome/Font-Awesome/blob/master/css/fontawesome.css) :thinking:

azefiel: Please use [recommended rules from Josh](https://joshsanders.github.io/telicon/#get-started) as we want icons to be 16px width/height by default (at least for the telicon collection).

If you feel the need to create one sheet per collection, that's fine.

src/js/lib/monster.ui.js
```diff
@@ -275,6 +275,28 @@ define(function(require) {
 				return monster.ui.getTemplatePhoneNumber(number.toString());
 			},
 
+			renderSvgIcon: function(id, cssClass) {
```
azefiel: We don't need to specify the action as this is a Handlebars helper.
```suggestion
			svgIcon: function(id, cssClass) {
```

src/js/lib/monster.ui.js
```diff
@@ -275,6 +275,28 @@ define(function(require) {
...
+			svgIcon: function(id, options) {
+				var template = monster.ui.getSvgIconTemplate({
+					id: id,
+					attributes: options.hash
+				});
+
+				return new Handlebars.SafeString(template);
+			},
+
+			telicon: function(id, options) {
```
azefiel: Variable declarations should be at the top of function blocks (only trumped by guards). Here, there is two ways to clean that up:
```javascript
telicon: function(pId, options) {
	var id = _.startsWith(pId, 'telicon2--')
			? pId
			: 'telicon2--' + pId,
		template = monster.ui.getSvgIconTemplate({
			id: id,
			attributes: options.hash
		});

	return new Handlebars.SafeString(template);
},
```
Or, more condensed:
```javascript
telicon: function(id, options) {
	return new Handlebars.SafeString(
		monster.ui.getSvgIconTemplate({
			id: _.startsWith(id, 'telicon2--')
				? id
				: 'telicon2--' + id,
			attributes: options.hash
		})
	);
},
```

src/js/lib/monster.ui.js
```diff
@@ -3093,6 +3115,43 @@ define(function(require) {
...
+			throw TypeError('"id" is not a string');
+		}
+		if (_.has(args, 'attributes') && !_.isPlainObject(args.attributes)) {
+			throw TypeError('"attributes" is not a plain object');
+		}
+		var iconId = args.id;
+		var iconPrefix = iconId.substring(0, iconId.indexOf('--'));
+		var attributes = _.get(args, 'attributes', {});
+		attributes.class
+			= _.chain(attributes)
```
azefiel: Please avoid linebreaks before or after `=` in an assignment.

Regarding lodash chaining, you can either keep `_.chain()` on the assignment line or just `_` and move `.chain()` on its own line.

src/js/lib/monster.ui.js
```diff
@@ -3093,6 +3115,43 @@ define(function(require) {
...
+				.get('class', '')
+				.split(' ')
+				.union([ 'svg-icon', iconPrefix ])
+				.uniq()
```
azefiel: `lodash#union` already checks for uniqueness so this operation is overkill.
```suggestion
```

src/js/lib/monster.ui.js
```diff
@@ -3093,6 +3115,43 @@ define(function(require) {
...
+				.join(' ')
+				.value();
+
+		var templateData = {
```
azefiel: No need of this variable, you can assign the data directly in the template request.

src/js/lib/monster.ui.js
```diff
@@ -3134,7 +3134,10 @@ define(function(require) {
 		attributes.class = _
 			.chain(attributes)
 			.get('class', '')
-			.split(' ')
+			.split(/\s+/g)	// Split by one or more whitespaces
+			.filter(function(className) {
```
azefiel: Nice addition to trim extra empty spaces and check against empty strings.

We could use a declarative style to make it more concise AND understandable!

To do that, you can do the opposite operation: reject empty strings instead of keeping non empty ones.
```javascript
_.reject(['my-css-class', ''], _.isEmpty)
```
This should give you the exact same output at the end of you chain.

guillegr123: :open_mouth: nice! It's shorter. I'll use that. thanks :+1:

src/js/lib/monster.ui.js
```diff
@@ -3134,11 +3134,10 @@ define(function(require) {
...
 			.chain(attributes)
 			.get('class', '')
-			.split(/\s+/g)	// Split by one or more whitespaces
-			.filter(function(className) {
-				return className.length > 0;	// Filter empty strings due to leading or trailing whitespaces, or empty string
-			})
+			.split(/\s+/g)      // Split by one or more whitespaces
+			.reject(_.isEmpty)  // Reject empty strings that appear due to leading or trailing whitespaces, or empty string
 			.union(['svg-icon', iconPrefix])
+			.uniq()
```
azefiel: By definition, `lodash#union` already guards against duplicates so this extra operation seems overkill.

guillegr123: Oh, that's true! Sorry, I missed that on Lodash's docs.

PR #279, opened 2019-02-25
UI-3313: Avoid getNumberFeatures() failure when number is mobile or is being ported

guillegr123:
Numbers with `state === 'port_in'` or mobile numbers do not have the properties `features_available` nor `_read_only.features_available`. However, they both have the property `features`, so an extra check was added for this one.

### DIFF
--- a/src/apps/common/submodules/numbers/numbers.js
+++ b/src/apps/common/submodules/numbers/numbers.js
@@ -26,6 +26,10 @@ define(function(require) {
 			var self = this;
 
 			_.each(arrayNumbers, function(number) {
+				if (number.state === 'port_in' || number.used_by === 'mobile') {
+					return;
+				}
+
 				var numberDiv = parent.find('[data-phonenumber="' + number.phoneNumber + '"]'),
 					args = {
 						target: numberDiv.find('.number-options'),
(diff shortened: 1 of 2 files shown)


guillegr123:
That sounds better. I added the guard for `mobile`, and also for `state === 'port_in'`, as those numbers in porting process do not contain the `available_features` property, and the features menu is not rendered for them either.

src/apps/common/submodules/numbers/numbers.js
```diff
@@ -26,6 +26,10 @@ define(function(require) {
 			var self = this;
 
 			_.each(arrayNumbers, function(number) {
+				if (number.state === 'port_in' || number.used_by === 'mobile') {
```
azefiel: Looks good!

src/js/lib/monster.util.js
```diff
@@ -1171,7 +1171,7 @@ define(function(require) {
 		if (!_.isPlainObject(number)) {
 			throw new TypeError('"number" is not an object');
 		}
-		if (!_.has(number, 'features_available') && !_.has(number, '_read_only.features_available')) {
+		if (!_.has(number, 'phoneNumber') || (!_.has(number, 'features_available') && !_.has(number, '_read_only.features_available'))) {
```
azefiel: Not a fan of checking against `phoneNumber` as that prop is added by the `numbers` common control and is not present on a regular number object.

```suggestion
		if (!_.has(number, 'features_available') && !_.has(number, '_read_only.features_available')) {
```

Objectively, the phone number object should not be muted but you shouldn't care about it a the moment.

guillegr123:
Done! Sorry for introducing the bug :sweat_smile:

PR #281, opened 2019-02-28
UI-3311: Remove e911 from features array when removing e911 data from number

guillegr123:


azefiel:
We shouldn't have to do this as Kazoo updates that list on it's own when `e911` is missing on phone number update.

guillegr123:
I thought I saw a time where it did not happen and Kazoo returned the same feature array after the POST request. However I tried this again today, and it did not happen, so I will be closing this PR.

PR #287, opened 2019-03-05
UI-3321: Pull alerts on login, non-core app load and account change

guillegr123:


guillegr123:
Yup, that is the reason.

PR #288, opened 2019-03-05
UI-3315:Port request listing

masmerino13:


src/apps/common/submodules/portListing/portListing.js
```diff
@@ -96,9 +105,17 @@ define(function(require) {
...
+					{
+						tabs: self.portListingCheckProfile() ? subTabs : tabs
+					}
+				]
+			});
+
 			if (isMonsterApp) {
 				parent = args.parent;
-				container = args.container;
+				container = $('#common_app_container .app-content-wrapper');
```
guillegr123: I think that container could change depending on the app where this component is rendered.

src/apps/common/submodules/portListing/views/ports-listing.html
```diff
@@ -47,11 +53,10 @@
...
 			{{/if}}
-				{{!-- <td data-filter-value="{{state}}{{#if extra.isScheduledToday}} scheduledtoday{{/if}}" data-sort-value="{{ tryI18n @root.i18n.portListing.misc.status state }}{{#compare state '===' 'scheduled'}}-{{scheduled_date}}{{/compare}}"> --}}
 				<td data-filter-value="{{state}}" data-sort-value="{{ tryI18n @root.i18n.portListing.misc.status state }}">
 				{{#compare state '===' 'submitted'}}
 					<div class="status-wrapper">
-						<i class="icon-telicon-prioritize"></i>
+						{{telicon "list-prioritize"}}
 						<div class="content">
 							<div class="status">
 								{{@root.i18n.portListing.misc.status.submitted}}
```
guillegr123: Here you could use `{{tryI18n @root.i18n.portListing.misc.status status}}`, and for the other statuses as well, to not to depend on the `{{#compare...`, as a first step to remove repeated tags. I wonder if something similar could be done for the icons. If not, then the `{{#compare...` could be kept only to render the icons, instead of each content div.

masmerino13: Thanks, but i'm not sure if this is really useful on this case

PR #289, opened 2019-03-06
UI-3319: Main topbar icon to display notifications count

guillegr123:


guillegr123:
Ready to review :slightly_smiling_face: 

Changes:
- Topbar element will be rendered without alerts before loading alerts from API, then it will be re-rendered when alerts have been received (commit [fb0f87b](https://github.com/2600hz/monster-ui/pull/289/commits/fb0f87b9cdf9691ebe8033c49f643f8c415615d0)).
- An additional flag named `preventRequestEvents` was added for the function `callApi`, which will cause that the events `monster.requestStart` and `monster.requestEnd` not to be raised (commit [46b3faf](https://github.com/2600hz/monster-ui/pull/289/commits/46b3faf32ac1e3123f45624c8f3b2094514435b9)) These are the events that the core app uses to display and hide the loader on the upper right side, per each request. [This flag was set to the API call for alerts](https://github.com/2600hz/monster-ui/blob/fb0f87b9cdf9691ebe8033c49f643f8c415615d0/src/apps/core/submodules/alerts/alerts.js#L137), to prevent to display the loader for that specific API request.

src/apps/core/submodules/alerts/alerts.js
```diff
@@ -16,10 +16,47 @@ define(function(require) {
...
 		alertsRender: function() {
-			var self = this;
+			var self = this,
+				initTemplate = function initTemplate(alerts) {
+					var alertGroups = alerts ? self.alertsFormatData({ data: alerts }) : {},
+						alertsCount = _.reduce(alertGroups, function(count, alertGroup) {
+							return count + alertGroup.length;
+						}, 0),
+						dataTemplate = {
+							alertsCount: alertsCount === 0 ? null : alertsCount > 9 ? '9+' : alertsCount.toString()
```
azefiel: Ternaries should not be nested, as it make its harder to understand the ongoing logic.

You could use two different props, one to determine if the alert count and one to contain the actual alert count.

src/apps/core/submodules/alerts/alerts.js
```diff
@@ -29,8 +66,32 @@ define(function(require) {
...
+		 * Bind template content events
+		 * @param  {Object} args
+		 * @param  {jQuery} args.template  Template to bind
+		 */
+		alertsBindEvents: function(args) {
+			var self = this,
+				$template = args.template;
+
+			$template.find('#main_topbar_alert_toggle_link').on('click', function() {
+				$(this).find('.badge').fadeOut({
```
azefiel: We generally use the [more concise](https://devdocs.io/jquery/fadeout#fadeOut-duration-complete) form of `jQuery#fadeOut`

src/apps/core/submodules/alerts/views/nav.html
```diff
@@ -0,0 +1,8 @@
+<li id="main_topbar_alert">
+	<a href="javascript:void(0);" class="links" id="main_topbar_alert_toggle_link" data-original-title="{{i18n.alerts.alertToggle.toggleTooltip}}" data-placement="bottom" data-toggle="tooltip">
```
azefiel:
```suggestion
	<a href="javascript:void(0);" class="links" id="main_topbar_alert_toggle_link" data-title="{{i18n.alerts.alertToggle.toggleTooltip}}" data-placement="bottom" data-toggle="tooltip">
```

src/css/partials/_base.scss
```diff
@@ -22,6 +23,9 @@ $french-gray: #c0c0c9;
 $mischka: #e0e0e9;
 $white-lilac: #F0F0F9;
 
+// Darks
+$woodsmoke: #151517;
```
azefiel: Please try to keep this list sorted, from darkest to lightest.

src/js/lib/monster.apps.js
```diff
@@ -328,9 +328,11 @@ define(function() {
 							apiSettings.headers['X-Kazoo-Cluster-ID'] = monster.config.kazooClusterId;
 						}
 
-						if (params.hasOwnProperty('onChargesCancelled')) {
-							apiSettings.onChargesCancelled = params.onChargesCancelled;
-						}
+						_.merge(apiSettings,
+							_.pick(params, [
+								'onChargesCancelled',
+								'preventRequestEvents'
```
azefiel: Mmmm, not a fan of this name. Maybe `bypassSpinner` would me more descriptive?

src/js/lib/monster.js
```diff
@@ -555,10 +555,14 @@ define(function(require) {
 			self.kazooSdk = $.getKazooSdk({
 				apiRoot: monster.config.api.default,
 				onRequestStart: function(request, requestOptions) {
-					monster.pub('monster.requestStart');
+					if (!_.get(requestOptions, 'preventRequestEvents', false)) {
```
azefiel: So, this works but it completely disables the requestStart/requestEnd events, which could be used for something else.

In fact, it is used by `insertTemplate()` to determine if a request is currently running.

A better way to do this would be to keep the event publishing but disable the spinner.

src/js/lib/monster.apps.js
```diff
@@ -311,7 +311,7 @@ define(function() {
 					if (cancelCall) {
 						return errorCallback && errorCallback();
 					} else {
-						var apiSettings = $.extend({
+						var apiSettings = _.assignIn({
```
guillegr123: I used [`lodash#assignIn`](https://lodash.com/docs/4.17.11#assignIn) because [it was not a deep copy](https://api.jquery.com/jquery.extend/), in order to not change the current behavior (just in case).

guillegr123:
Ready to review again :sweat_smile:

src/apps/core/app.js
```diff
@@ -651,9 +655,15 @@ define(function(require) {
...
 
-		onRequestStart: function($spinner) {
+		onRequestStart: function(args) {
 			var self = this,
-				waitTime = 250;
+				waitTime = 250,
+				$spinner = args.spinner;
+
+			// If indicated, bypass progress indicator display/hide process
+			if (_.get(args, 'bypassProgressIndicator', false)) {
```
azefiel: You should unpack `args` parameters in their own variable. It makes the code easier to read and also forces you to set default values at the same time/location (on variable declaration).

```suggestion
			if (isBypassingProgressIndicator) {
```

src/apps/core/app.js
```diff
@@ -675,9 +685,15 @@ define(function(require) {
...
 
-		onRequestEnd: function($spinner) {
+		onRequestEnd: function(args) {
 			var self = this,
-				waitTime = 50;
+				waitTime = 50,
+				$spinner = args.spinner;
+
+			// If indicated, bypass progress indicator display/hide process
+			if (_.get(args, 'bypassProgressIndicator', false)) {
```
azefiel: Same here.

src/apps/core/submodules/alerts/alerts.js
```diff
@@ -16,10 +16,48 @@ define(function(require) {
...
+						$template = initTemplate(alerts);
+
+					if ($topbarAlert.length === 0) {
+						$template.insertBefore($navLinks.find('#main_topbar_signout'));
+					} else {
+						$topbarAlert.replaceWith($template);
+					}
+				};
 
 			monster.waterfall([
```
azefiel: `monster.ui@waterfall` is not needed here since your functions do not require results from the previous ones.
```suggestion
			monster.series([
```

src/js/lib/monster.apps.js
azefiel: 👍 

Maybe leave a comment explaining why `lodash#assignIn` is used here since we will usually prefer using `merge`/`assign`.

guillegr123: :+1: I'll add it

azefiel:
Ah, I should have mentioned, a new PR would have been ideal for that documentation work because now the notification related work is going to be stalled until that page is finished...

guillegr123:
Doc changes reverted and moved to [OOB-callApi-documentation](https://github.com/2600hz/monster-ui/pull/312).

PR #295, opened 2019-03-13
UI-3320: Notifications dropdown

guillegr123:


src/apps/core/submodules/alerts/alerts.scss
```diff
@@ -0,0 +1,157 @@
...
+						margin-right: 0;
+						margin-bottom: 0;
+						color: $white;
+						font-size: 12px;
+						font-weight: normal;
+						line-height: 12px;
+					}
+
+					.icon {
+						float: left;
```
azefiel: Have you heard about this cool thing called `flexbox` ;)

Try using instead of floating elements. You might need to add an extra container that will contain the content of your alert. Again, better semantic structure with clear class name that are self explanatory in that context is what we aim for here.

guillegr123:
Sorry, I noticed after requesting a new review that I forgot to push the code changes. I just pushed them, so it should be good for review now.

PR #298, opened 2019-03-14
UI-3316: Getting port request if port authority

masmerino13:


src/apps/common/submodules/portListing/portListing.js
```diff
@@ -246,7 +246,8 @@ define(function(require) {
 							},
 						numbers: numbers,
 						numbersAmount: _.size(numbers),
-						timeline: self.portListingFormatEntriesToTimeline(results.timeline)
+						timeline: self.portListingFormatEntriesToTimeline(results.timeline),
+						allowPrivate: monster.util.isSuperDuper() || monster.util.isPortAuthority(_.get(portRequest, 'port_authority', null))
```
azefiel: 👍

src/js/lib/monster.util.js
```diff
@@ -263,6 +263,13 @@ define(function(require) {
 			return isSuperDuper;
 		},
 
+		// Validate if the given port authority ID matches with the current account ID
+		isPortAuthority: function(portAuthorityID, pAccount) {
```
azefiel: This shouldn't a global util as it is specific to port request, which are handled by the porting app. Please keep it contained in the relevant port common control.

src/js/lib/monster.util.js
```diff
@@ -263,6 +263,13 @@ define(function(require) {
...
+			var account = pAccount || (monster.apps.hasOwnProperty('auth') && monster.apps.auth.hasOwnProperty('originalAccount') ? monster.apps.auth.originalAccount : {});
```
azefiel: Once you move back that method in the relevant common control, you should simply check against `self.accountId` since you won't be able to see the related tab while masquerading.

src/apps/common/submodules/portListing/portListing.js
```diff
@@ -999,6 +1000,22 @@ define(function(require) {
 			});
 		},
 
+		portListingRequestByAgent: function(args) {
```
azefiel: Try to keep methods with `Request` in their name for methods containing actual request calls (e.i. with either `callApi()` or `monster.request()`).

src/apps/common/submodules/portListing/portListing.js
```diff
@@ -1030,14 +1047,16 @@ define(function(require) {
...
 			switch (args.tab) {
-				case 'descendants':
-					self.portlistingRequestDescendantsByType(args);
-					break;
 				case 'account':
-				case 'agent':
 				default:
 					self.portListingRequestByType(args);
 					break;
+				case 'agent':
```
azefiel: By convention, the `default` case should always be the last one.

src/js/lib/monster.util.js
masmerino13: seems i don't really needs this method, just compare `_.get(portRequest, 'port_authority', null) === self.accountId` on https://github.com/2600hz/monster-ui/pull/298#discussion-diff-270157809R250

PR #306, opened 2019-03-25
UI-3335: Telicon icons update to v2.2.0 + markdown fixes

guillegr123:
Took the chance to apply some fixes to `handlebars.md` file:
- Change nested list indent from 2 to 4 spaces to display the nested list properly, [due to changes in Python Markdown to comply with standards](https://github.com/mkdocs/mkdocs/issues/545).
- Add missing `#` symbol to `compare` example.


azefiel:
Please fix the reference doc page `SVG Icons` as well: https://docs.2600hz.com/ui/docs/svgIcons/

Most likely a missing empty line before the list starts.

guillegr123:
Fixed, it was a missing break and spacing.

docs/svgIcons.md
```diff
@@ -2,25 +2,26 @@ title: SVG Icons
 
 # SVG Icons in Monster-UI
 Monster UI includes 4 SVG icon collections:
+
 1. Google Drive Logos
```
azefiel: I'd argue this list does not need to be ordered. In fact, it should not be a list but rather titles:

```suggestion
##### Google Drive Logos
```

docs/svgIcons.md
```diff
@@ -2,25 +2,26 @@ title: SVG Icons
...
+    - `sso-google--color`: Google color
+    - `sso-google--grey`: Google grey
+    - `sso-office365--color`: Office365 color
+    - `sso-office365--grey`: Office365 grey
+    - `sso-salesforce--color`: Salesforce color
+    - `sso-salesforce--grey`: Salesforce grey
+    - `sso-linkedin--color`: LinkedIn color
+    - `sso-linkedin--grey`: LinkedIn grey
+4. [Telicon Icons v2][telicon]  
+A third party SVG icon system for Monster UI. The list of icons is available on [the official website][telicon2iconlist]. To use these icons in Monster UI,you can use the dedicated `telicon` Handlebars helper: `{{telicon "phone-outbound"}}`.
```
azefiel:
```suggestion
A third party SVG icon system for Monster UI. The list of icons is available on [the official website][telicon2iconlist]. To use these icons in Monster UI, you can use the dedicated `telicon` Handlebars helper: `{{telicon "phone-outbound"}}`.
```

src/apps/core/views/svg-container.html
```diff
@@ -468,557 +468,703 @@
 		<!-- | ============================ -->
 		<!-- |
 			 |  [1] - Account tree
```
azefiel: This numbered list is a nice addition but I believe this is not included in the library.

Updating telicon to a new version should be fairly simple and one should not have to regenerate that list every time. Plus, there is already a way lookup existing icons from the github page.

guillegr123: That is true, it takes too much time to update the list and the numbering for each symbol. I'll remove them.

PR #310, opened 2019-03-28
UI-3328: Add voicemail.patch method in kazooSdk doc

guillegr123:


guillegr123:
Oh, Ok, good to know.

PR #318, opened 2019-04-03
UI-3320: Store a local (ephemeral) list of read/dismissed notifications for rendering purposes

azefiel:


azefiel:
@guillegr123 

Please review each commit chronologically so it makes more sense. The goal here was to implement a poor man's cache system for read/dismissed notification's that will expire on logout.

src/apps/core/submodules/alerts/alerts.js
```diff
@@ -258,6 +305,9 @@ define(function(require) {
 				}).map(function(alerts, type) {
 					return {
 						type: type,
+						unreadCount: _.reduce(alerts, function(total, alert) {
```
guillegr123: Just found that `lodash#sumBy` can be used in this case, to make it a little bit shorter:

```
unreadCount: _.sumBy(alerts, function(alert) {
    return _.includes(readAlertIds, alert.id) ? 0 : 1;
}),
```

src/apps/core/submodules/alerts/alerts.js
```diff
@@ -193,55 +193,51 @@ define(function(require) {
...
-						title: alert.title,
-						metadata: metadata,
-						message: alert.message,
-						category: category,
-						clearable: alert.clearable
-					};
+					return _.merge({
+						metadata: _
+							.chain(alert.metadata)
+							.reject(function(value, key) {
```
guillegr123: Here `lodash#reject` turns the original metadata object into an array, so the key for the following `lodash#map` function is the array index, instead of the object key.

guillegr123:
The declarative style looks nice, it is cleaner and shorter :smile:

src/apps/core/submodules/alerts/alerts.js
```diff
@@ -262,7 +261,7 @@ define(function(require) {
 						alerts: alerts
 					};
 				}).sortBy(function(alertGroup) {
-					return _.get(sortOrder, alertGroup.type) + alertGroup.type;
+					return _.includes(sortOrder, alertGroup.type) ? _.indexOf(sortOrder, alertGroup.type) : _.size(sortOrder);
```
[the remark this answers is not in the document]
azefiel: It will do for now.

This will have to be refactored along the `groupBy` once we have to handle notifications from more apps.

src/apps/core/submodules/alerts/alerts.js
azefiel: Good catch, I'll apply the reject once the object is transformed into an array.

PR #319, opened 2019-04-04
[master] QF: UI-3219 Bug introduced in #230

bputtick:
Bug introduced in #230 Typo for caller/callee in balance.js


azefiel:
Congrats 🎉 🎉 🎉 

Thanks again for fixing that bug!

(This new feature should be packaged into the next release)

PR #326, opened 2019-04-11
UI-3361: Drag and drop media upload

masmerino13:


src/apps/common/submodules/dragableUploads/dragableUploads.js
```diff
@@ -0,0 +1,216 @@
...
+		},
+
+		appFlags: {
+			dragableUploads: {
+				files: [],
+				maxFileSize: 5242880 // 5MB
+			}
+		},
+
+		renderUploadArea: function(args) {
```
azefiel: Don't forget to append the submodule's name to each of its methods.

src/apps/common/submodules/dragableUploads/dragableUploads.scss
```diff
@@ -0,0 +1,160 @@
+@import '../../../../css/partials/base';
+
+.dragable-wrapper {
```
azefiel: Please use a more specific selector here, usually it should contain the name of the common control.

src/apps/common/submodules/dragableUploads/dragableUploads.js
masmerino13: oops, thanks!

PR #327, opened 2019-04-11
UI-3139: Normalize path format for all platforms

azefiel:
Format all paths in unix style (backslash) since it is supported by all
platforms.

requirejs and chokidar (used by gulp.watch) expects unix formatted paths
and silently fails when a Windows path is provided.

This change impacts the whole build system and will hopfully avoid path
related error later on.

### DIFF
--- a/gulp/helpers/helpers.js
+++ b/gulp/helpers/helpers.js
@@ -1,4 +1,4 @@
-import { join } from 'path';
+import { join } from 'upath';
 import fs from 'fs';
 import parser from 'yargs-parser';
 import { src } from '../paths.js';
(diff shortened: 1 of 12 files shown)


azefiel:
Tested all tasks on both Windows 10 and Ubuntu 18.04 (WSL).

guillegr123:
Tested on Windows 10, Node v10.15.3, NPM v6.4.1, and tasks work OK, but watches are not triggered when running on dev mode.

azefiel:
Good catch, it should at least work with the current LTS version. I'll do more testing.

PR #328, opened 2019-04-11
UI-3342: Update monsterPanelText layout and standardize warning colors for form inputs

guillegr123:


src/apps/core/views/monster-panel-text-danger.html
```diff
@@ -1,6 +1,6 @@
 <div class="monster-panel-text {{#if className}} {{className}}{{/if}}" data-type="danger">
 	<div class="monster-panel-text-header clearfix">
-		<div class="icon-wrapper"><i class="fa fa-exclamation-circle"></i></div>
+		<div class="icon-wrapper"><i class="fa fa-exclamation-circle icon"></i></div>
```
azefiel: 2 things:

- try not using the `icon` class as it might inherit styling from bootstrap and telicon style sheets (I personally use `iconography` instead but you can use anything)
- wy not use the `warning-triangle` icon from telicon2?

src/css/lib/monster.scss
```diff
@@ -685,80 +685,79 @@ label[class*='monster-radio'].prepend-label > span {
...
-	box-sizing: border-box;
-	border-left-width: 4px;
-	border-left-style: solid;
+	display: block;
+	margin-bottom: 1rem;
 	width: 320px;
-}
+	padding: 0;
+	background: $panel-background-color;
+	border-radius: 2px 2px 2px 2px;
```
azefiel: When you have the same value for all four directions, repeating it is superfluous.
```suggestion
	border-radius: 2px;
```

src/apps/core/views/monster-panel-text-danger.html
guillegr123: OK. About the `warning-triangle`, I did not use that one because it is the same for warning. But on a second thought, it does not matter a lot because it will have a different color.

src/css/lib/monster.scss
guillegr123: Did not noticed it 😅 Thanks!

src/apps/core/views/monster-panel-text-danger.html
azefiel: Maybe check with Josh before making that change. It does seem like telicon2 does not have an `error` specific icon.

guillegr123: Will use `x--circle` as suggested by Josh.

PR #341, opened 2019-04-24
UI-3372: Stop to send unnecessary request when filter by date ranges

masmerino13:


src/apps/common/submodules/portListing/portListing.js
```diff
@@ -504,6 +507,25 @@ define(function(require) {
...
+
+						self.portListingRenderTableList({
+							template: self.portListingGet('container'),
+							fromDate: fromDate,
+							toDate: toDate,
+							type: type,
+							tab: tab
+						});
+					});
+
```
azefiel: Curious why you moved this? AFAICT there is no changes?

It is okay move code around but you should create a specific commit with a message explaining your intent so PR reviews do not turn into "where is waldo" puzzles.

masmerino13: This is what was causing the weired issue, this event was binding with the table pagination/rendering causing the extra click events, but this element (filter button) isn't part of the table, this is why I moved it to the correct binding method.

My bad, I will create a different commit for future cases, thanks for the suggestion

PR #349, opened 2019-05-03
MSTR-1: Tag selector common control

masmerino13:


masmerino13:
Since this will be a common control that will be used in different modules, it will be nice to have your comments and suggestions @azefiel

azefiel:
Just stumbled upon this http://sliptree.github.io/bootstrap-tokenfield/ and I wouldn't be against wrapping it in a `monster.ui` helper vs writing a common control as it implements all the features necessary for a tag selector.

masmerino13:
Yeah, looks good to me, should this one works with monster UI Boostrap version?
cc: @azefiel

azefiel:
I'd say don't import the library's CSS and instead, see with Josh what style this new component should take (be sure to ask the behavior for every state: disabled, validation, is there a specific behavior on hover, active etc).

masmerino13:
Hey, I have noticed (thanks to Pilar) `Chosen` already have multiple selection that basically works like the same.

I think we can have a common control that receive array of tags and just personalize the default `chosen` styles?

cc: @azefiel

azefiel:
`chosen` is just a fancy wrapper over `select` and only allows you to selected values from a predetermined list of `option` tags: you cannot add new options, which is kind the whole point of a tag selector.

azefiel:
And as previously mentioned, this feature should be implemented as a UI helper (`monster.ui` method) rather than a full blown common control since it does not contain asynchronous operations.

azefiel:
@masmerino13 

What was the decision on this, do we need a full-blown tag selector or using `chosen` with multi-selecting enabled enough for our purposes?

azefiel:
@masmerino13 ping ^^

masmerino13:
ups, sorry for the delay,

Doing a quick search, I found `chosen` can be customized as `text free` for multiple selections
http://jsfiddle.net/masmerino/etcb9n47/2/

What do you think about it?

cc: @azefiel

masmerino13:
So, I think `chosen` do the job

azefiel:
I thought you were supposed to talk to Mark and/or Josh to determine the requirements of this feature so we could decide which option was best.

masmerino13:
Not yet,

I didn't talk to @mark2600 or Josh about this feature, we are just using chosen.

azefiel:
@masmerino13 

Closing since we don't need this feature anymore.

PR #353, opened 2019-05-07
UI-3380: Show agent tab only if SDA or self managing ports

guillegr123:


src/apps/common/submodules/portListing/portListing.js
```diff
@@ -100,15 +100,21 @@ define(function(require) {
...
-						text: self.i18n.active().portListing.tabs[tab],
-						id: tab,
-						menus: [{
-							tabs: subTabs
-						}]
-					};
-				}),
+				tabs = _
+					.chain(self.appFlags.portListing.tabs)
+					.filter(function(tab) {
```
azefiel: This could be switched to use `lodash#reject` for clarity, especially when ready the condition.

src/apps/common/submodules/portListing/portListing.js
```diff
@@ -100,15 +100,21 @@ define(function(require) {
...
+						return tab !== 'agent' || monster.util.isSuperDuper() || monster.config.whitelabel.port.authority === self.accountId;
```
azefiel: That whole path won't necessarily be set (`monster.config.whitelable` will for sure).

src/apps/common/submodules/portListing/portListing.js
guillegr123: Sounds good, the intent will be more clear with `lodash#reject`.

PR #354, opened 2019-05-08
MSTR-5: Filter completed port requests by modified date

guillegr123:


src/apps/common/submodules/portListing/portListing.js
```diff
@@ -1168,25 +1168,31 @@ define(function(require) {
...
 		/**
+		 * @param  {Object}   args
 		 * @param  {Function} args.success
-		 * @param  {String} args.tab
-		 * @param  {String} args.status
-		 * @param  {Object} args.filters
+		 * @param  {String}   args.tab
+		 * @param  {String}   args.type
+		 * @param  {Date}     [args.fromDate]
+		 * @param  {Date}     [args.toDate]
```
azefiel: Imagine later on you have to add a new parameter that can be either a string or a number. The type would be defined as `{String|Number}`, which would be longer than `{Function}`, in which case, by following your logic, you would have to update the spacing of every single parameter listed, relative to their type.

I am not saying this is bad per se but it sure looks like it would add a lot of overhead when the original intent was to simply document a new parameter.

The take away is, you shouldn't care too much about keeping all parameters aligned.

PR #357, opened 2019-05-15
MACF-2: Add key-value editor UI helper

guillegr123:


docs/monster-ui/keyValueEditor().md
```diff
@@ -0,0 +1,76 @@
...
+    addLink: self.i18n.active().customData.addLink,
+    keyPlaceholder: self.i18n.active().customData.keyPlaceholder,
+    valuePlaceholder: self.i18n.active().customData.valuePlaceholder
+  }
+};
+
+monster.ui.keyValueEditor(target, options);
+```
+
+[append]: http://api.jquery.com/append/
```
azefiel:
```suggestion
```

src/apps/core/views/monster-key-value-editor-row.html
```diff
@@ -0,0 +1,9 @@
+<div class="monster-key-value-editor-row">
+	<div class="data-key">
+		<input required id="{{inputName}}.key" name="{{inputName}}.key" type="text" placeholder="{{coalesce keyPlaceholder i18n.keyValueEditor.keyPlaceholder}}" value="{{key}}"></input>
+	</div>
+	<div class="data-value">
+		<input required id="{{inputName}}.value" name="{{inputName}}.value" type="text" placeholder="{{coalesce valuePlaceholder i18n.keyValueEditor.valuePlaceholder}}" value="{{value}}"></input>
+	</div>
+	<i class="fa fa-trash-o fa-lg monster-red" data-action="remove"></i>
```
azefiel: Actionable icons should be wrapped into `a` tags for accessibility purposes.

Also, maybe use a `telicon@2` icon here.

The `data-action` seems overkill, you can use a class on the `a` tag once implemented.

Lastly, the repetition of those icons looks heavy so having them defined with `visibility: hidden` by default and only shown on row hover should be easier on the eye.

src/apps/core/views/monster-key-value-editor.html
```diff
@@ -0,0 +1,4 @@
+<div class="monster-key-value-editor">
+	<div class="monster-key-value-data-container"></div>
+	<a class="monster-link" data-action="add">{{coalesce addLink i18n.keyValueEditor.addRow}}</a>
```
azefiel:
```suggestion
	<a class="monster-link" data-action="add">
		{{coalesce addLink i18n.keyValueEditor.addRow}}
	</a>
```
Same here, the `data-action` is not necessary and and `id` tag can be used instead for better selector targeting.

src/css/lib/monster.ui.scss
```diff
@@ -944,3 +944,37 @@ i.monster-password-strength {
...
+			i[data-action="remove"] {
+				cursor: pointer;
+				flex: initial;
+				padding-top: 4px;
+			}
+		}
+	}
+
+	.monster-link[data-action="add"] {
+		width: 180px;
```
azefiel: Not necessary as this element is displayed inline.

src/js/lib/monster.ui.js
```diff
@@ -3215,21 +3215,119 @@ define(function(require) {
...
+		}
+		if (!_.isPlainObject(source)) {
+			throw TypeError('"source" is not a plain object');
+		}
+
+		return _.mergeWith(object, source, function(objValue, srcValue, key) {
+			objValue = _.isNil(objValue) ? '' : objValue;
+			srcValue = _.isNil(srcValue) ? '' : srcValue;
+			if (!_.isString(objValue || '') || !_.isString(srcValue || '')) {
+				throw TypeError('One or both values is not a string');
```
azefiel: This seems too restrictive, you should try casting non-string types into strings rather than raising an error.

Also, errors should be instantiated with the `new` keyword.

docs/monster-ui/keyValueEditor().md
```diff
@@ -0,0 +1,76 @@
...
+Key | Description | Type | Default | Required
+:-: | --- | :-: | :-: | :-:
+`$target` | A jQuery object inside which the key-value editor will be inserted using the [append()][append] method. | `jQuery` | | `true`
+`options` | A plain JavaScript object that contains options to set up the editor. | `Object`([#options](#options)) | | `false`
+
+#### options
+Key | Description | Type | Default | Required
+:-: | --- | :-: | :-: | :-:
+`data` | JSON object that contains the data to be handled by the editor. | `Object` | `{}` | `false`
+`inputName` | String to be set as prefix for the text fields' ID and name attributes. | `String` | `data` | `false`
```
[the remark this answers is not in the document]
guillegr123: I removed the `ID` attributes, but added the description as an extra paragraph to describe how the widget interacts with `monster.ui.getFormData()`.

src/css/lib/monster.ui.scss
```diff
@@ -944,3 +944,48 @@ i.monster-password-strength {
...
+					height: 30px;
+				}
+			}
+
+			.key-value-remove {
+				color: $font-primary-color;
+				cursor: pointer;
+				padding-top: 0.2rem;
+				opacity: 0;
+				transition: visibility 0.25s, opacity 0.25s ease-in-out;
```
azefiel: Why transition the visibility as well as the opacity?

I suggest you pick only one property: for this use case, `visibility` seems like the appropriate choice (don't forget to update the property on hover).

Also, the transition speed seems a bit slow, 1/8 of a second should keep thinks snappy.

src/js/lib/monster.ui.js
```diff
@@ -3215,21 +3215,119 @@ define(function(require) {
...
+				value: value,
+				inputName: inputName + '[' + index + ']',
+				keyPlaceholder: keyPlaceholder,
+				valuePlaceholder: valuePlaceholder
+			}));
+		};
+		// Add initial rows
+		var counter = 0;
+		_.each(data, function(value, key, index) {
+			addRow(value, key, counter++);
```
azefiel: Using unary increment/decrement operators is usually not a good idea as it can hide unintentional effects leading to unexpected behaviors.

It is more expressive to increment/decrement variables by using statements:
```suggestion
			addRow(value, key, counter);
			counter += 1;
```

src/js/lib/monster.ui.js
```diff
@@ -3215,21 +3215,119 @@ define(function(require) {
...
+		});
+		// Bind events
+		$editorTemplate.find('.key-value-add').on('click', function(e) {
+			e.preventDefault();
+			addRow('', '', counter++);
+		});
+		$rowContainer.on('click', '.key-value-remove', function(e) {
+			e.preventDefault();
+			$(this).closest('.monster-key-value-editor-row').remove();
```
azefiel: Maybe you can leave a comment explaining why `counter` is not decremented on remove (`getFormData()` takes care of fixing the indexing).

src/js/lib/monster.ui.js
```diff
@@ -3215,21 +3215,119 @@ define(function(require) {
...
+		return $editorTemplate;
+	}
+
+	/**
+	 * Merges HTML attributes, mapped as JSON objects
+	 * @param   {Object} object  Destination object
+	 * @param   {Object} source  Source object
+	 * @returns {Object}         Returns `object` after merge
+	 */
+	function mergeHtmlAttributes(object, source) {
```
azefiel: Not sure how this relates to the task at hand, it does not seem like this function is used by `monster.ui#keyValueEditor` in any way.

azefiel:
Looks good, some minor changes required before merging though.

src/js/lib/monster.ui.js
azefiel: To clarify, this does not need to be remove, just wondering why it was implemented alongside the feature since it is not related in any way.

src/css/lib/monster.ui.scss
guillegr123: You're right, one seems to be enough. I'll leave opacity, because it adds a fade effect, and visibility is only a blink.

src/js/lib/monster.ui.js
guillegr123: I added the function because initially I thought to also have a Handlebars helper, and wanted to allow the user to add more attributes to the HTML tag like it is done with the icon helpers. However, late in the development I dismissed the idea, because the editor was going to require some interactions via JQuery, and the Handlebars helpers only return plain text, not JQuery objects. I decided to leave the function there because it can be used for other Handlebars helpers in the future.

PR #358, opened 2019-05-16
MSTR-8: Group timeline entries by day with time zone

guillegr123:


docs/monster-util/getCurrentTimeZone().md
```diff
@@ -0,0 +1,23 @@
...
+## Syntax
+```javascript
+monster.util.getCurrentTimeZone();
+```
+
+### Return value
+A `String` identifier for the current time zone.
+
+## Description
+This method gets the identifier of the current time zone.
```
azefiel:
```suggestion
The `monster.util.getCurrentTimezone` method returns a [TZ database name](https://www.wikiwand.com/en/Tz_database) representing the timezone set for the currently authenticated session.
```

src/js/lib/monster.util.js
```diff
@@ -1012,6 +1012,20 @@ define(function(require) {
 		return formatter.format(base).replace('NaN', '');
 	}
 
+	/**
+	 * Gets the current time zone identifier.
```
azefiel:
```suggestion
	 * Returns the timezone of the currently authenticated session
```

guillegr123:
Thanks, it was by chance, because our time zone is slightly different than the default one in Sandbox.

PR #366, opened 2019-05-29
MSTR-7: App selector common control

guillegr123:
Please check commit 5a9e25e, which has another way to retrieve the selected applications from the app selector, and allows to have multiple instances of the control active. The final implementation, added in commit 75ecdcd, only allows one instance of the control at a time, as the selected values are global to the app.


src/apps/common/submodules/appSelector/appSelector.js
```diff
@@ -0,0 +1,377 @@
...
+						},
+						$template = $(self.getTemplate({
+							name: 'layout',
+							data: dataTemplate,
+							submodule: 'appSelector'
+						})),
+						$appList = $template.find('.app-list');
+
+					_.each(selectedApps, function(app) {
+						$appList.find('[data-id="' + app.id + '"]').addClass('selected');
```
azefiel: This should be done in the template directly, as you already iterate through that list.

You can pass an array of selected app ids and use `Handlebars#ifInArray` to conditionally add the `selected` class.
```handlebars
<ul class="app-list">
{{#each apps}}
	<li class="app-item text-container{{#each tags}} {{this}}{{/each}}{{#ifInArray id @root.selectedAppIds}} selected{{/ifInArray}}" data-id="{{id}}" data-name="{{toLowerCase label}}">
		<div class="app-icon {{extraCssClass}}" {{#if icon}}style="background-image:url({{icon}})"{{/if}}></div>
		<div class="app-title">
			{{label}}
		</div>
		<div class="app-description">
			{{description}}
		</div>
	</li>
{{/each}}
</ul>
```

src/apps/common/submodules/appSelector/appSelector.js
```diff
@@ -0,0 +1,377 @@
...
+				selectedAppIds = _.get(args, 'selectedAppIds', []),
+				$container = args.container,
+				initTemplate = function initTemplate(apps) {
+					var selectedApps = self.appSelectorGetStore('selectedApps', []),
+						tagCount = _.chain(apps).flatMap('tags').countBy().value(),
+						filters = _.transform(self.appFlags.appSelector.tagFilters, function(filters, filterName) {
+							filters[filterName] = (filterName === 'all') ? _.size(apps) : _.get(tagCount, filterName, 0);
+						}, {}),
+						dataTemplate = {
+							apps: apps,
```
azefiel: Looks like this is an object as well? An array would be better, especially for sorting purposes.

src/apps/common/submodules/appSelector/appSelector.js
```diff
@@ -0,0 +1,377 @@
...
+
+					monster.ui.tooltips($template, {
+						selector: '.selected-title .selected-text',
+						options: {
+							placement: 'bottom',
+							title: function() {
+								return _
+									.chain(self.appSelectorGetStore('selectedApps', []))
+									.map('label')
+									.join(', ')
```
azefiel: It makes it easier to parse visually when alphabetically sorted:
```suggestion
									.sortBy()
									.join(', ')
```

guillegr123:
Thanks! No, I have not tried fadein/fadeout. You mean the methods from jQuery, right?

About `getAppList`, I was thinking of a case where you want to display the icon and names of the selected apps on the caller side. It would be easier to have all the application data to do that, without having to do any further requests or process. However, I agree that a simple array of app ids selected seems cleaner. Do you think that the case I mention could happen, and if it will be worthy to have all the app metadata? Other option, more specialized, could be to pass to the function a string array that contains the list of metadata properties that we want to get from the method, and if none is provided, then only the array of app IDs is returned.

guillegr123:
I have tried using jQuery's `fadeIn` and `fadeOut`, but it removes the sliding effect of the list to the left and to the right, because they set a fixed width in the icon container element style. Because of it, I opted to animate the white shadow instead, to expand and fade-in when adding an app icon to the list, and to decrease and fade-out on removal.

azefiel:
> Thanks! No, I have not tried fadein/fadeout. You mean the methods from jQuery, right?
> 
> About `getAppList`, I was thinking of a case where you want to display the icon and names of the selected apps on the caller side. It would be easier to have all the application data to do that, without having to do any further requests or process. However, I agree that a simple array of app ids selected seems cleaner. Do you think that the case I mention could happen, and if it will be worthy to have all the app metadata? Other option, more specialized, could be to pass to the function a string array that contains the list of metadata properties that we want to get from the method, and if none is provided, then only the array of app IDs is returned.

I guess we could have a utility method to abstract app metadata retrieval that could select the language to use by default but getting that metadata from a list of ids is as trivial as:
```javascript
_
	.chain(monster.appsStore)
	.values()
	.filter(function(app) {
		return _.includes(appIds, app.id);
	})
	.value();
```
I don't think it should be the job of this common control to provide that metadata though.

src/apps/common/submodules/appSelector/appSelector.js
```diff
@@ -0,0 +1,387 @@
...
+							selectedApps: self.appSelectorGetAppsByIds(selectedAppIds)
+						},
+						$template = $(self.getTemplate({
+							name: 'layout',
+							data: dataTemplate,
+							submodule: 'appSelector'
+						})),
+						$appList = $template.find('.app-list');
+
+					_.each(selectedAppIds, function(appId) {
```
azefiel: This loop should not be necessary after https://github.com/2600hz/monster-ui/pull/366#discussion_r289581864

guillegr123: Sorry, I ended up forgetting to remove this.

PR #368, opened 2019-06-03
MSTR-10: Create Handlebars helper to lookup an object path

guillegr123:


guillegr123:
I wanted to use Handlebars partials, to avoid repeating code to generate the field prompts in the form. I only pass the path to the partial, and use it as a name to assign to the input element, and use the helper to lookup for the value.

PR #371, opened 2019-06-07
MSTR-15: Customize navigation wizard

masmerino13:
support custom actions and title

### DIFF
--- a/src/apps/common/submodules/navigationWizard/navigationWizard.scss
+++ b/src/apps/common/submodules/navigationWizard/navigationWizard.scss
@@ -5,9 +5,19 @@
 }
 
 #navigation_wizard_wrapper .header {
-	padding: 15px 0 15px 30px;
+	display: flex;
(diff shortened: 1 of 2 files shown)


src/apps/common/submodules/navigationWizard/navigationWizard.scss
```diff
@@ -109,19 +112,10 @@
...
 
 #navigation_wizard_wrapper #cancel {
-	float: right;
-	bottom: 38px;
-	right: 25px;
 	min-width: 72px;
 	width: 72px;
 }
 
-#navigation_wizard_wrapper #cancel, #navigation_wizard_wrapper .back {
```
pcandia: by removing this now the cancel button and back are just plain gray, I think you should be this section

masmerino13: you are the best, thanks!

PR #372, opened 2019-06-07
MSTR-16: New param to return file list without upload them

masmerino13:
Return files without upload them


src/apps/common/submodules/dragableUploads/dragableUploads.js
```diff
@@ -181,6 +184,8 @@ define(function(require) {
 		dragableUploadsRequestMediaUpload: function(callback, file) {
 			var self = this;
 
+			console.log('file', file);
```
pcandia: You forgot to remove the console.log line

masmerino13: ouch, thanks!

PR #378, opened 2019-06-14
MSTR-24: Do not throw error if alerts template is not set

guillegr123:


guillegr123:
I've checked the code and ran some tests, and verified that if the `/alerts` request fails, the top icon and dropdown indicates that there are no notifications.

Also, there is no race condition, as the alerts component is rendered empty while requesting the alerts list to the API and retrieving the UI restrictions. Only when all three processes have finished, the alerts component is rendered again with the retrieved alerts. The advantage of this is that the alerts component will be displayed (empty, though) while still retrieving the data from the API. The disadvantage is to render the component twice.

guillegr123:
Is it OK this way, or do I render the alerts component only once, when the request to the API has finished?

azefiel:
Keep it the way it is, it looks better from a UX standpoint.

You still should use `generateErrors: false` for `/alerts` so the end-user is not randomly prompted with a error dialog whenever the request fails (500/502 for example).

Just call your success callback in the error handler as well, and pass down an empty array.

guillegr123:
Ready. The empty array is passed to the parallel callback at `alertsRender` instead of  the error callback `alertsRequestListAlerts` because that function is related to the rendering of the control, whereas `alertsRequestListAlerts` is an utility function, which could be applied to another context (although it is not the case at the time).

src/apps/core/submodules/alerts/alerts.js
```diff
@@ -187,11 +187,9 @@ define(function(require) {
 			var self = this,
 				$template = self.appFlags.alerts.template;
 
-			if (!$template) {
-				throw new ReferenceError('The notifications template has not been loaded.');
+			if ($template instanceof $) {
```
azefiel: You could reverse this check and use it as a guard.

PR #386, opened 2019-07-01
MSTR-30: Add spinner utility function to UI lib

guillegr123:


guillegr123:
It is for the trunk limits in [the restrictions step at the account wizard](https://share.goabstract.com/1b0f86c1-ae03-4f4c-a301-d4f565fa966a?collectionLayerId=f74fdce5-bcd9-4f69-bf10-0932df43cc56&mode=design). `<input type="number"/>` could be used, but the look and feel of jQuery UI's spinner is more similar to other UI controls

guillegr123:
I'll close this PR, because I need to adjust some CSS styles. However, let me know @azefiel if you are OK to add this widget, or to use the `number` input instead.

guillegr123:
I reopened this PR because the new changes were kept in the original branch.

src/js/lib/monster.ui.js
```diff
@@ -3490,6 +3522,55 @@ define(function(require) {
...
+	 * Transforms a field into a jQuery UI Spinner element
+	 * @param  {jQuery} $target  Input to transform
+	 * @param  {Object} options  List of options
+	 * @return {jQuery}          jQuery UI Spinner instance
+	 */
+	function spinner($target, pOptions) {
+		if (!($target instanceof $)) {
+			throw TypeError('"$target" is not a jQuery object');
+		}
+		if (!_.isPlainObject(pOptions)) {
```
azefiel: Is `pOptions` mandatory or optional? I see it is named `pOption` so assume its optional but the jsdoc does not reflect that (`options` is not between `[]`).

If it is optional, you should update your check to see if it is not both `undefined` and a plain object at the same time.

src/css/lib/monster.ui.scss
```diff
@@ -992,3 +995,42 @@ i.monster-password-strength {
 		font-size: 0.75rem;
 	}
 }
+
+/************************************************************/
+/********************* JQUERY UI WIDGETS ********************/
+/************************************************************/
+.ui-widget.ui-spinner {
```
azefiel: jQuery UI widgets overrides should be in `monster-ui\src\css\lib\custom\jquery-ui-add.scss`

docs/monster-ui/spinner().md
```diff
@@ -0,0 +1,31 @@
...
+## Syntax
+```javascript
+monster.ui.spinner(target[, options]);
+```
+
+### Parameters
+Key | Description | Type | Default | Required
+:-: | --- | :-: | :-: | :-:
+`$target` | Input element on which the method will be applied. | `jQuery` | | `true`
+`options` | List of options for [jQuery UI Spinner Widget][spinner]. | `Object` | | `false`
```
azefiel: You should notify which props are not overridable.

guillegr123: Do you mean the props for `options`? If so, all of them are overridable, because any of the options for jQuery UI Spinner widget can be set.

guillegr123:
Maybe `numberPicker` or `numberSpinner` instead of just `spinner`?

azefiel:
> `numberPicker`

That one seems more descriptive don't you think?

docs/monster-ui/spinner().md
azefiel: I believe `change` ([line](https://github.com/2600hz/monster-ui/pull/386/files#diff-602a66ca60a9e72a9cdcf92ae9ec90cdR3543)) isn't, no?

guillegr123:
Yes, I think so. I'll use it then.

docs/monster-ui/spinner().md
guillegr123: The default implementation set in the helper is not overridable itself, but a `change` prop can still be passed, and it will be called at the end of the default `change` function.

azefiel: Ah, you might document that behavior then.

guillegr123: Yes, I added a description for it in the markdown document at 719dcd7

I described what the control does inside the `change` function, though I did not mention the `change` prop or event directly. Is it OK?

azefiel: You could me more specific and add an extra `options.change` entry in the parameters' table with a description such as:

> Callback on value change (note that this function if specified will not override the widget's behavior but can be used to retrieve the new value).

guillegr123: Done :+1:

PR #391, opened 2019-07-12
MSTR-37: Show or exclude specific apps by its IDs

guillegr123:


src/apps/common/submodules/appSelector/appSelector.js
```diff
@@ -137,6 +147,17 @@ define(function(require) {
 					self.appSelectorSetStore('apps', _.keyBy(appList, 'id'));
 					self.appSelectorSetStore('selectedAppIds', selectedAppIds);
 
+					// Filter apps
+					if (!_.isEmpty(showOnlyAppIds)) {
```
azefiel: This check seems too strict, you should be able to specify an empty array of ids (btw, do you know if the common control has an empty state? I know it was not part of the mockup/spec).

Here, we need to check if the `showOnlyAppIds` is not undefined and an array:
```suggestion
					if (
						!_.isUndefined(showOnlyAppIds)
						&& _.isArray(showOnlyAppIds)
					) {
```
And simply use the default value set by `lodash#get` when declaring `showOnlyAppIds` (which is `undefined`).

src/apps/common/submodules/appSelector/appSelector.js
```diff
@@ -137,6 +147,17 @@ define(function(require) {
...
+						appList = _.filter(appList, function(app) {
+							return _.includes(showOnlyAppIds, app.id);
+						});
+					} else if (!_.isEmpty(excludeAppIds)) {
```
azefiel: Same here.

src/apps/common/submodules/appSelector/appSelector.js
```diff
@@ -72,14 +72,24 @@ define(function(require) {
 		 * @param  {Object} args
 		 * @param  {jQuery} args.container  Element that will contain the app selector
 		 * @param  {('all'|'account'|'user')} [args.scope='all'] App list scope
-		 * @param  {Boolean} [args.forceFetch=false]  Force to fetch app data from API instead of using the cached one
+		 * @param  {Boolean} [args.forceFetch=false]  Force to fetch app data from API instead of
+		 *                                            using the cached one
+		 * @param  {String[]} [args.showOnlyAppIds=[]]  List of IDs for the specific apps to be
```
azefiel:
```suggestion
		 * @param  {String[]} [args.availableApps=[]]  List of IDs for the specific apps to be
```

src/apps/common/submodules/appSelector/appSelector.js
```diff
@@ -72,14 +72,24 @@ define(function(require) {
...
+		 *                                              displayed. This is applied on top of the
+		 *                                              selected scope, and takes precedence over
+		 *                                              args.excludeAppIds.
+		 * @param  {String[]} [args.excludeAppIds=[]]  List of App IDs of the apps to be excluded
```
azefiel:
```suggestion
		 * @param  {String[]} [args.excludedApps=[]]  List of App IDs of the apps to be excluded
```

src/apps/common/submodules/appSelector/appSelector.js
guillegr123: The common control does not have an empty state, but it will render empty if no apps are listed, due to no existing apps for the specified scope or due to the restrictions.

azefiel: Okay, I'll create a ticket and ping Josh about it.

PR #392, opened 2019-07-15
MSTR-1: Allow chosen to support text free tags

masmerino13:


src/js/lib/monster.ui.js
```diff
@@ -475,9 +475,34 @@ define(function(require) {
...
+				}, pOptions),
+				chosen;
 
 			$target.chosen(options);
+
+			if (!pOptions.tags) {
+				return;
+			}
+
+			chosen = $target.data('chosen');
```
guillegr123: I think it would be better to change the name of this variable, to not to cause any confusion with the method name when a dev scans the code quickly.

azefiel: Agree, since `data('chosen')` effectively returns the Chosen instance, you can rename this variable to `instance`.

src/js/lib/monster.ui.js
```diff
@@ -475,9 +475,34 @@ define(function(require) {
...
+					return;
+				}
+
+				var newTag = $(this).val(),
+					option = $('<option>').val(newTag).text(newTag);
+
+				// Add the new option
+				$target.prepend(option);
+				// Automatically select it
+				$target.find(option).prop('selected', true);
```
azefiel: A more straightforward way of doing this it to set the `select` value directly:
```suggestion
				$target.val(newTag);
```

src/js/lib/monster.ui.js
```diff
@@ -475,9 +475,34 @@ define(function(require) {
...
+
+			if (!pOptions.tags) {
+				return;
+			}
+
+			chosen = $target.data('chosen');
+
+			// Bind the keyup event to the search box input
+			chosen.container.find('input').on('keyup', function(e) {
+				// If we hit Enter and the results list is empty (no matches) add the option
```
azefiel: 👍

src/js/lib/monster.ui.js
masmerino13: awesome, than yuo guys

src/js/lib/monster.ui.js
masmerino13: Looking trough the code this line isn't required, I just add the `prop('selected', true)` on the `option` generation

https://github.com/2600hz/monster-ui/pull/392/files#diff-602a66ca60a9e72a9cdcf92ae9ec90cdR500

src/js/lib/monster.ui.js
```diff
@@ -494,12 +494,13 @@ define(function(require) {
...
 
 				var newTag = $(this).val(),
-					option = $('<option>').val(newTag).text(newTag);
+					option = $('<option>')
+						.val(newTag)
+						.text(newTag)
+						.prop('selected', true);
 
 				// Add the new option
 				$target.prepend(option);
```
azefiel: Seems like using `jquery#prepend` messes up the number of options selected:
1. add a new option by hitting enter
1. select a different option than the one just added
1. click on chosen to open the dropdown

=> the option selected in the dropdown is still the one added programmatically

Looks like adding the option with `jquery#append` does not produce this side effect but please test more use cases involving chosen's different options (multilpe selection for example).

masmerino13: awesome, let me try this

masmerino13:
Thank you @azefiel , `jquery#append` works like a charm!

I already did those changes and also I did some test just to make sure this works for different scenarios.

PR #398, opened 2019-07-23
MPOR-6: Allow to update port req on submit error

guillegr123:
The change is to keep the requestId after save, before status change. This will allow to know that the port request is already saved in case the status change fails, in order to update it on further attempts instead of creating a new one.

### DIFF
--- a/src/apps/common/submodules/portWizard/portWizard.js
+++ b/src/apps/common/submodules/portWizard/portWizard.js
@@ -1321,35 +1321,60 @@ define(function(require) {
 					.on('click', function(event) {
 						event.preventDefault();
 
-						self.portWizardHelperSavePort(_.merge({}, args, {
-							stopErrorPropagation: true,
-							success: function(requestId) {
+						monster.waterfall([
+							function(waterfallCallback) {
+								self.portWizardHelperSavePort(_.merge({}, args, {
+									stopErrorPropagation: true,
+									success: function(requestId) {
+										waterfallCallback(null, requestId);
+									}
+								}));
+							},
+							function(requestId, waterfallCallback) {
+								// Add request ID to args object, to know that the request
+								// has already been saved at this point
+								args = _.merge({
+									data: {
+										request: {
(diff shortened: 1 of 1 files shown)


src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -1321,35 +1321,60 @@ define(function(require) {
...
+									stopErrorPropagation: true,
+									success: function(requestId) {
+										waterfallCallback(null, requestId);
+									}
+								}));
+							},
+							function(requestId, waterfallCallback) {
+								// Add request ID to args object, to know that the request
+								// has already been saved at this point
+								args = _.merge({
```
azefiel: You should use `lodash#set` to assign a value to that property as it takes care of creating those nested objects if missing.

src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -1475,6 +1500,7 @@ define(function(require) {
 							data: self.portWizardNormalizePortRequestData(args.data.request)
 						},
 						success: function(portRequest) {
+							_.set(args, 'data.request.id', portRequest.id);
```
azefiel: I believe there is a store for this, so maybe you should save the whole payload:
```suggestion
							self.portWizardSet('portRequest', portRequest);
```

azefiel: Never mind, doesn't seem like the wizard's steps use the store.

guillegr123: Yeah, it does not seem to use it.

src/apps/common/submodules/portWizard/portWizard.js
guillegr123: I used `lodash#merge` to not to modify the original `args` object, but I think that it does not really matter, as it probably is modified along the way anyways. I'll change it to `lodash#set`.

PR #415, opened 2019-08-12
Update Readme

jon-braz:
Added a quick explanation of some of the `gulp` commands available

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,9 +47,9 @@ define({
 
 *For a full list and comprehensive descriptions of all the configurable options, head over to the [dedicated documentation](https://docs.2600hz.com/ui/docs/configuration/)*.
 
-### Build
+### Build/Serve (locally)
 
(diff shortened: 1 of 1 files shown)


azefiel:
Hey @jon-braz 

I believe the page you are looking for is [this one](https://docs.2600hz.com/ui/docs/gulpCommand/)? (also accessible on GitHub [here](https://github.com/2600hz/monster-ui/blob/master/docs/gulpCommand.md))

It should be referenced in the readme under the "Build" section but let me check real quick.

azefiel:
Ah, I see. Some extra information was added to the note explaining where to find documentation about our building tasks.

I'll update that to clarify it and hopefully it will be enough.

azefiel:
How does [that look like](https://github.com/2600hz/monster-ui/blob/master/README.md#build), any better?

PR #418, opened 2019-08-13
MSTR-59: Make phoneNumber rule to ignore empty values

guillegr123:


src/js/lib/monster.ui.js
```diff
@@ -1349,7 +1349,7 @@ define(function(require) {
 			});
 
 			$.validator.addMethod('phoneNumber', function(value, element) {
-				return monster.util.getFormatPhoneNumber(value).isValid;
+				return _.isEmpty(value) || monster.util.getFormatPhoneNumber(value).isValid;
```
azefiel: You should use the built-in `$.validator#optional` method as it will check the field's state:
```suggestion
				return this.optional(element) || monster.util.getFormatPhoneNumber(value).isValid;
```

guillegr123: Thanks! Did not know about that method.

azefiel: Yeah, I obviously didn't know about it either when implementing this validation rule.

PR #438, opened 2019-09-04
MSTR-74: Lowercase username on login

danielfinke:
Since the username is normalized to lowercase when a user is created in Kazoo `cb_users`, it should also be possible to authenticate with any case.


azefiel:
4.3 -> https://github.com/2600hz/monster-ui/pull/457

PR #441, opened 2019-09-09
MSTR-67: Scroll to top invalid field on validation

guillegr123:


guillegr123:
Thanks! :smile: They offer something to [focus the first or last focused invalid field](https://jqueryvalidation.org/validate/#focusinvalid), but it has to be explicitly invoked after validation, and does not have a scroll transition.

PR #451, opened 2019-09-13
MSTR-68: Create `formatNumber` utility function

azefiel:


azefiel:
@guillegr123 

Please try to review this before I get Monday in if you have time.

I still need to improve the doc and deprecate `monster.util#formatPrice`.

src/js/lib/monster.util.js
```diff
@@ -889,6 +889,48 @@ define(function(require) {
 	}
 	util.formatMacAddress = formatMacAddress;
 
+	/**
+	 * Wrapper over Intl.NumberFormat constructor's format function
+	 * @param  {Object} args
+	 * @param  {Number|String} args.number Number to format.
+	 * @param  {Number} [args.digits] Number of fractional digits to use.
+	 * @param  {'currency'|'decimal'|'percent'} [args.digits='decimal'] Formating style to use.
```
guillegr123:
```suggestion
	 * @param  {('currency'|'decimal'|'percent')} [args.style='decimal'] Formating style to use.
```

azefiel:
> I just have a question regarding the behavior for `currency` style. When the number of decimal `digits` is lower than the number of minor unit digits for the currency. it is ignored, and the minimum digits is used. [...]
> 
> Is this the expected behavior?

Yes it is expected as the `digits` prop matches `minimumFractionDigits` from the `Intl.NumberFormat` constructor. The explanation of the the `digits` prop is misleading as-is.

In your example, it will return a result with at least one fractional digit, and up to `maximumFractionDigits` (also defined on instance creation).

We don't really want to expose those other properties, at least for `currency`, as the constructor uses defaults relative to the locale used. It might be needed to expose them for `decimal|percent` but we'll address that if/when that need comes.

docs/monster-util/formatNumber().md
```diff
@@ -0,0 +1,150 @@
...
+monster.util.formatNumber(args);
+```
+
+### Parameters
+`args` is a mandatory `Object` parameter with the following properties:
+
+Key | Description | Type | Default | Required
+--- | --- | --- | --- | ---
+`number` | Number to format. | `Number`, `String` | | `true`
+`digits` | Minimum number of fractional digits (the default for plain number and percent formatting is 0; the default for currency formatting is the number of minor unit digits provided by the [ISO 4217 currency code list](https://www.currency-iso.org/en/home/tables/table-a1.html) (2 if the list doesn't provide that information).). | `Number` | | `false`
```
guillegr123: The double parenthesis at the end of the description looks a little bit odd.

docs/monster-util/formatNumber().md
```diff
@@ -0,0 +1,150 @@
...
+  number: 5,
+  digits: 3
+});
+// output: "5.000"
+
+monster.util.formatNumber({
+  number: 5.5,
+  digits: 3
+});
+// output: "5.556"
```
guillegr123:
```suggestion
// output: "5.500"
```

docs/monster-util/formatNumber().md
```diff
@@ -0,0 +1,150 @@
...
+  digits: 3
+});
+// output: "$5.000"
+
+monster.util.formatNumber({
+  style: 'currency',
+  number: 5.5,
+  digits: 3
+});
+// output: "$5.556"
```
guillegr123:
```suggestion
// output: "$5.500"
```

azefiel:
Mmmm, it's not being ignored since it is used to determine the minimum fraction digits and not the maximum.

I'll think over it more and see if settings both `minimumFractionDigits` and `maximumFractionDigits` makes sense for all use cases.

guillegr123:
Sorry for the force-push above, I accidentally committed and pushed changes in this branch instead of mine, so I deleted them to return the history as it was before my mistake.

Commit history: https://github.com/2600hz/monster-ui/commits/MSTR-68

docs/monster-util/formatNumber().md
```diff
@@ -0,0 +1,66 @@
...
+### Format as plain number
+```javascript
+monster.util.formatNumber({ number: 5 });                   // -> "5"
+monster.util.formatNumber({ style: 'decimal', number: 5 }); // -> "5"
+monster.util.formatNumber({ number: -5 });                  // -> "-5"
+monster.util.formatNumber({ number: 5000 });                // -> "5,000"
+monster.util.formatNumber({ number: '5.5' });               // -> "5.5"
+monster.util.formatNumber({ number: 5.4321 });              // -> "5.432"
+monster.util.formatNumber({ number: 5, digits: 3 });        // -> "5.000"
+monster.util.formatNumber({ number: 5.55, digits: 1 });     // -> "-5.6"
```
guillegr123: Jus a small typo in the result, or the `number`:

```suggestion
monster.util.formatNumber({ number: 5.55, digits: 1 });     // -> "5.6"
```

docs/monster-util/formatNumber().md
```diff
@@ -0,0 +1,66 @@
...
+```
+### Format as percentage
+```javascript
+monster.util.formatNumber({ style: 'percent', number: .5 });                // -> "50%"
+monster.util.formatNumber({ style: 'percent', number: '.5' });              // -> "50%"
+monster.util.formatNumber({ style: 'percent', number: 0.5 });               // -> "50%"
+monster.util.formatNumber({ style: 'percent', number: -5, digits: 2 });     // -> "-50.00%"
+monster.util.formatNumber({ style: 'percent', number: .555 });              // -> "56%"
+monster.util.formatNumber({ style: 'percent', number: .12345, digits: 2 }); // -> "12.35%"
+monster.util.formatNumber({ style: 'percent', number: .5, digits: 3 });     // -> ".5%"
```
guillegr123: Here the result is not the actual one. The output would be: `"50.000%"`

PR #453, opened 2019-09-17
MSTR-73: Fix and update styles for common controls

guillegr123:
* Update checkbox design.
* Align bullet at navigation wizard nav menu.
* Disable text selection at app selector
* Fix header shadow on navigation wizard title bar over right content panel.
* Fix app logo misalign after resize.
* Update primary color. Also use variables whenever possible.


src/css/partials/_base.scss
```diff
@@ -1,6 +1,6 @@
 /* _colors.scss */
 /* unique color names */
-$dodger-blue: #22A5FF;
+$dodger-blue: #2297FF;
```
azefiel: I hope whoever asked you to update this understands it will impact all apps.

guillegr123: I'll ask Bret and Josh about it. Bret pointed out the differences in color in some of the common controls like the navigationWizard and the checkboxes, so I checked the mockups for the accounts wizard and the app exchange submission and noticed that the color referred as primary in the CSS of the common controls was `#2297FF` in the mockups, instead of `#22A5FF`. So I decided to update it as it seemed that it was the new primary color and the difference to the old one was slight. I also wanted to know your opinion about it when you reviewed this PR.

image

I made the color change in a single commit in case it needs to be reverted.

guillegr123: I asked to Bret and Josh, and they agreed :+1: `#22A5FF` was the old color, and `#2297FF` is the current primary color.

azefiel: Okay, as long as they understand the whole UI will be impacted, even older mockups that where not designed specifically with that color).

PR #455, opened 2019-09-17
MSTR-72: Ask for confirmation before wizard unload

guillegr123:


src/apps/common/submodules/navigationWizard/navigationWizard.js
```diff
@@ -13,7 +13,7 @@ define(function(require) {
 		appFlags: {
 			// Default values just for reference, they will be overwritten anyways
 			navigationWizard: {
-				askForConfirmationOnCancel: false,
+				askForConfirmationBeforeLeave: false,
```
azefiel:
```suggestion
				askForConfirmationBeforeExit: false,
```

src/apps/common/submodules/navigationWizard/navigationWizard.js
```diff
@@ -118,6 +119,12 @@ define(function(require) {
 				stepId: navigationWizardFlags.currentStep
 			});
 
+			if (navigationWizardFlags.askForConfirmationBeforeLeave) {
+				$(window).on('beforeunload', function() {
```
azefiel: When binding events on `window` or any top DOM element, you should always think about unbinding them whenever possible.

guillegr123: I forgot about it, thanks :+1:

PR #458, opened 2019-09-25
MSTR-78: Scroll through popup rather than body when selecting models

azefiel:
Following the behaavioral update of dialogs, scrolling on the body was
disabled when a dialog is open.

Here we had to update the scrolling target when a model is selected
since the original one was body and could break the UX.

### DIFF
--- a/src/apps/common/submodules/chooseModel/chooseModel.js
+++ b/src/apps/common/submodules/chooseModel/chooseModel.js
@@ -233,7 +233,9 @@ define(function(require) {
 							submodule: 'chooseModel'
 						}));
 				templateDevice.find('.block-footer').slideDown(function() {
-					$('html, body').animate({ scrollTop: templateDevice.find('div.block-device-info div.title-bar').offset().top }, function() {
+					popup.animate({
+						scrollTop: popup.offset().top
+					}, 250, function() {
 						templateDevice.find('#name').focus();
 					});
 				});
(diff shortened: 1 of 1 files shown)


src/apps/common/submodules/chooseModel/chooseModel.js
```diff
@@ -233,7 +233,9 @@ define(function(require) {
 							submodule: 'chooseModel'
 						}));
 				templateDevice.find('.block-footer').slideDown(function() {
-					$('html, body').animate({ scrollTop: templateDevice.find('div.block-device-info div.title-bar').offset().top }, function() {
+					popup.animate({
+						scrollTop: popup.offset().top
```
guillegr123: It seems that `popup.offset().top` is the position of the top of the dialog from the viewport's top, so the scrolling is always 49px. However, the text field is focused at the end of the animation, so it gets on screen anyways.

Maybe scroll to the bottom of the dialog instead?

```suggestion
						scrollTop: popup.get(0).scrollHeight
```

guillegr123: Also, this is out of the scope of this ticket, but I noticed that I needed to have the apps `desktop-app` and `callthru-us` to be able to load the SmartPBX app. Does that happen to you too?

azefiel: Ah weird, my original selector was `popup.find('.block-device-info')` to scroll to that section specifically, but it doesn't seem like it made it to the PR 😢

azefiel:
4.3 -> https://github.com/2600hz/monster-ui/pull/459

PR #460, opened 2019-09-27
MSTR-79: Guard against apps extending unpermitted apps

azefiel:


azefiel:
4.3 -> https://github.com/2600hz/monster-ui/pull/461

PR #462, opened 2019-10-01
MSTR-77: Hide elements to print wizard step only

guillegr123:


src/apps/common/submodules/navigationWizard/navigationWizard.scss
```diff
@@ -172,3 +172,41 @@
 	background: #2ED022;
 	color: #071F05;
 }
+
+@media print {
+	.core-topbar-wrapper {
+		display: none;
+	}
```
azefiel: I believe it would make more sense for this selector to be declared in `core/style/app.scss`, with a dedicated media query.

src/apps/common/submodules/navigationWizard/navigationWizard.scss
```diff
@@ -172,3 +172,41 @@
...
+	symbol[id^="telicon2"] {
+		&, * {
+			color: inherit !important;
+		}
+	}
+	*/
+
+	@page {
+		margin: 0.75in;
+	}
```
azefiel: Same here as it would be applied globally.

guillegr123:
4.3 -> #463

PR #464, opened 2019-10-04
OOB: Fix check of which listeners to keep

arjanmvos:
Hello 2600hz,

I came across a small bug in the monster.socket.js.
The check for which listeners to keep was looking in the wrong variable, which caused ALL listeners to be removed when unsubscribing from the binding.

This pull request fixes that. Would you be able to also merge this back into older versions?

Thank you!

Kind regards,

Arjan Vos
IP Telecom


arjanmvos:
Perfect, thank you @azefiel !

PR #469, opened 2019-10-16
SPRO-3: Support new Polycom models

vlad88sv:
VVX 101, 201
Trio 8500, 8800


azefiel:
> I think that the references for these new models need to be added to [src/css/lib/monster.ua.css](src/css/lib/monster.ua.css) and [src/js/lib/monster.ua.js](src/js/lib/monster.ua.js). Is that the case @azefiel ?

That would be nice, although I believe that module is not up-to-date with the current list of local model images.

PR #473, opened 2019-10-21
MSTR-88: Expose manual quantities post endpoint

azefiel:


azefiel:
4.3 -> https://github.com/2600hz/monster-ui/pull/474

PR #477, opened 2019-10-22
MSTR-82: Add country selector helper to UI library

guillegr123:


guillegr123:
4.3 -> #480

docs/monster-ui/countrySelector().md
```diff
@@ -0,0 +1,31 @@
+title: countrySelector()
+
+# monster.ui.countrySelector()
+
+## Syntax
+```javascript
+monster.ui.countrySelector($target[, selectedValues, options]);
```
azefiel: I'd rather you have `selectedValues` defined as part of `options` to avoid having to pass down `undefined` as a placeholder parameter:
```js
monster.ui.countrySelector($target, undefined, {...});
```
Something like `monster.ui.countrySelector($target, { selectedValues: [...], options: {...}})` looks more practical, let me know what you think.

src/apps/core/views/monster-country-selector-item.html
```diff
@@ -0,0 +1,5 @@
+<span class="monster-country-item-label">{{{label}}}</span>
+{{#if code}}
+	{{!-- A background image was used instead of the img element to avoid displaying the broken image placeholder for unexistent flag images --}}
+	<span class="monster-country-item-flag" style="background-image:url(css/assets/flags/24/{{code}}.png)"></span>
+{{/if}}
```
azefiel: Unless the mockup says otherwise, the flag should be place before the country name to keep them aligned (since they are the same size) vertically. It will make it much more easier to parse visually.

As for using `background-image` instead of an `img` tag, that works but you will most likely have to create a container for your flag with the same width/height and always displayed to act as a placeholder.

The second solution would be to have a default flag image (not sure if we have one) and use an `img` tag like so:
```html
<img src="path/to/original/image.png" onerror="path/to/fallback/image.png">
```

This can also be achieved with CSS only:
```html
<img src="path/to/original/image.png">
```
```scss
img {
	position: relative;

	&:before {
		content: ' ';
		display: block;
		position: absolute;
		top: 0;
		left: 0;
		height: 24px;
		width: 24px;
		background: transparent url('path/to/fallback/image.png');
	}
}
```

But then again, it all depends if we have a fallback image or not.

guillegr123:
> Curious if you'd considered using an existing [library](https://github.com/djgrant/chosen-image) instead of hacking chosen (I don't even know if it works, just wondering).

Yes, I was between that one, and [Image-Select](http://websemantics.github.io/Image-Select/). At the end I ended up mixing both codes :sweat_smile: The first one did not work for multi-select, and had a small bug where the image of the selected item was removed if the dropdown was opened and closed without changing the item.

The Image-Select plugin looks promising, although I had not tried it. I also was not sure to add another library script to the UI. Do you think it would be good to use that one instead of the custom code? I think it won't be hard to modify the code to use that plugin.

docs/monster-ui/countrySelector().md
guillegr123: I agree with you. Do you think it would be OK to include the `$target` parameter in that object too? Or is it better to leave it as a separated parameter, because it is required?

src/apps/core/views/monster-country-selector-item.html
guillegr123: Indeed, there is a fallback image :open_mouth: 

There is not a mockup for the actual dropdown list, only for the selected item, and the flag is placed to the right for the latter:

image

However I agree with you about displaying the image to the left. Do you think it would look odd if I display the flag to the left for the list items, and to the right for the selected item?

docs/monster-ui/countrySelector().md
azefiel: I'd leave it as its own parameter since it is required, plus it's a pattern we use throughout most of our UI helpers (the newer ones at least).

src/apps/core/views/monster-country-selector-item.html
azefiel: I think the flag should precede the country label in any case for visual consistency.

As you said, the mockup does not come with an expanded dropdown so I assume that use case was not taken into account during design.

guillegr123:
This PR was closed, because the branch will be rebased on top of master, to include the updated version of jQuery Chosen plugin.

PR #483, opened 2019-10-25
MSTR-104: Bump jQuery Chosen version to 1.8.7

guillegr123:
Tested some normal cases (single and multi-select) and custom`tags` option.


guillegr123:
4.3 -> #484

PR #485, opened 2019-10-28
MSTR-82: Add country selector helper to UI library

guillegr123:


guillegr123:
4.3 -> #486

src/js/lib/monster.ui.js
```diff
@@ -3354,9 +3356,62 @@ define(function(require) {
...
+		var selectedValues = _.get(args, 'selectedValues'),
+			options = _.get(args, 'options');
+		if (!_.isUndefined(selectedValues) && !_.isString(selectedValues) && !_.isArray(selectedValues)) {
+			throw TypeError('"args.selectedValues" is not a string nor an array');
+		}
+		if (!_.isUndefined(options) && !_.isPlainObject(options)) {
+			throw TypeError('"args.options" is not a plain object');
+		}
+
+		var selectedValues = _.isUndefined(selectedValues) ? [] : selectedValues,
```
azefiel: `selectedValues` is already declared above.

azefiel:
Regarding the missing flags, please update them as follows (as in, update the file not the reference code):
- BV -> NO
- GF -> FR
- GP -> FR
- HM -> AU
- PM -> FR
- SJ -> NO
- UM -> US

I checked real quick and all of those officially use the flag of the country they part of.

The only missing one now is IO (British Indian Ocean Territory) 🎉

guillegr123:
> Regarding the missing flags, please update them as follows (as in, update the file not the reference code)

About this, you mean creating a copy of the flag image file (for example, `FR.png`) and naming it with the code `GF.png`?

azefiel:
Yes, I wasn't sure if the missing flags were actually missing an or just referenced the placeholder flag.

guillegr123:
Thank you very much for your comments on variable hoisting and your suggestion to make the helper implementation more straightforward, I will keep them in mind :+1: I previously thought that the hoisting was only for the variable declaration, and did not include its initialization.

azefiel:
> Thank you very much for your comments on variable hoisting and your suggestion to make the helper implementation more straightforward, I will keep them in mind 👍 I previously thought that the hoisting was only for the variable declaration, and did not include its initialization.

It does if you initialize a variable during declaration.

PR #487, opened 2019-10-29
OOB: Handle PyYAML yaml.load(input) deprecation

azefiel:
The following warning was thrown on CI build:
> ./scripts/validate_mkdocs.py:40: YAMLLoadWarning: calling yaml.load_all() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.

Following the deprecation guide, this update simply specifies a `Loader=` parameter when loading the mkdoc toc.


azefiel:
4.3 -> https://github.com/2600hz/monster-ui/pull/488

PR #490, opened 2019-10-30
MSTR-91: Highlight and display select list error

guillegr123:


guillegr123:
I initially thought of adding the invalid styles to the `chosen-add.scss` file, so I started grouping the styles there. But at the end I considered that `monster.ui.scss` was the proper place for them. But the job was already done in `chosen-add.scss`, so I left these changes.

guillegr123:
4.3 -> #491

guillegr123:
Good to know, thanks for the heads up!

PR #498, opened 2019-10-31
MSTR-105: Import common styles from account wizard

guillegr123:


guillegr123:
Related -> [monster-ui-accounts#74](https://github.com/2600hz/monster-ui-accounts/pull/74)
4.3 -> #499

PR #500, opened 2019-11-04
Fix monster.util.friendlyTimer

arjanmvos:
friendlyTimer was not displaying hours if hours=0

monster.util.friendlyTimer(86461);
Should return: '01:00:01:01'
Instead it returned: '01:01:01';

This pull request solves this.

Also updated inline comment and fixed a spelling error :)
The comment has been updated as this function is now also used in other apps (such as CDRs, callqueues (qubicle), voicemails, and is used as a Handlebars Helper)


azefiel:
Hey @arjanmvos, thanks for you contribution!

Not sure where `Merge pull request #1 from 2600hz/master` ([this](https://github.com/2600hz/monster-ui/pull/1)?) comes from though. Would you mind cleaning that up or create a new PR base off an up-to-date version of master?

PR #518, opened 2019-11-25
MSTR-83: Create name and numbers port request step

guillegr123:


src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -173,25 +224,59 @@ define(function(require) {
...
+					.split(',')
+					.map(_.trim)
+					.reject(_.isEmpty)
+					.map(monster.util.getFormatPhoneNumber)
+					.uniqBy('e164Number')
+					.value();
+
+				nameAndNumbersData.numbersToPort.numbers = _
+					.chain(nameAndNumbersData.numbersToPort.formattedNumbers)
+					.map('originalNumber')
```
azefiel: What do you think of displaying numbers in their user format to have consistent formatting?

src/apps/common/submodules/portWizard/portWizard.scss
```diff
@@ -394,3 +394,15 @@
...
+	&#step_name_and_numbers {
+		#numbers_to_port_numbers {
+			min-height: 16rem;
+
+			&.monster-invalid + label.monster-invalid {
+				position: static;
+			}
+		}
+	}
+}
```
azefiel: Since you are starting from scratch on this project, maybe you could add a `partials` folder ([ex](https://github.com/2600hz/monster-ui-service-plan-override/tree/master/submodules/itemized)) and use one for each template to separate styles (you can also create a `common` one for styles reused by each step).

src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -157,12 +165,55 @@ define(function(require) {
...
+
+									return _
+										.chain(numbersText)
+										.split(',')
+										.map(_.trim)
+										.reject(_.isEmpty)
+										.map(monster.util.getFormatPhoneNumber)
+										.every('isValid')
+										.value();
+								}
```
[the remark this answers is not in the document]
guillegr123: Yeah, I was wondering if it was a good solution, because even the plugin does not provide this kind of rule. `normalizer` + `phoneNumber` sounds good, I'll do it this way.

src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -157,12 +165,55 @@ define(function(require) {
...
+							}
+						},
+						messages: {
+							'numbersToPort.numbers': {
+								custom: 'One or more numbers are invalid'
```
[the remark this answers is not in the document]
guillegr123: I totally forgot this string :sweat_smile:

src/apps/common/submodules/portWizard/portWizard.js
guillegr123: I would prefer it that way, but I think that some users may get confused if they see any of the numbers in a different format. They may feel that it has been changed. What do you think?
Similar to this, I was also considering to remove the repeated numbers.

src/apps/common/submodules/portWizard/portWizard.scss
guillegr123: Sounds good! Can you give me access to `monster-ui-service-plan-override`, so that I can see the example?

src/apps/common/submodules/portWizard/portWizard.js
azefiel: > I would prefer it that way, but I think that some users may get confused if they see any of the numbers in a different format. They may feel that it has been changed. What do you think?

Those numbers would be rendered using the user preferred format (hopefully) in the next step anyways so we might as well do it here.

My use case was if someone copy pasted number that would be in different format (international, national, e164) it will not look uniform.

Since the only reason a user would come back to this step would be to edit a number, maybe using `e164` would make more sense as it is easier to edit than in national format?

> Similar to this, I was also considering to remove the repeated numbers.

Yes, that would be expected, they can always review which numbers where added in the next step.

src/apps/common/submodules/portWizard/portWizard.scss
azefiel: Unfortunately I can't but it basically looks like this:

```
/moduleName
  /partials
    _common.scss
    _templateNameA.scss
    _templateNameB.scss
    ...
  /views
    _templateNameA.html
    _templateNameB.html
    ...
  moduleName.js
  moduleName.scss
```
And in `moduleName.scss` you simply import your partials nested under the cc root selector.

That way each template can have its own dedicated style sheet.

src/apps/common/submodules/portWizard/portWizard.js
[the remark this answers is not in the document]
azefiel: No worries, it happens to the best of us.

src/apps/common/submodules/portWizard/portWizard.scss
```diff
@@ -1,4 +1,5 @@
 @import '../../../../css/partials/base';
+@import 'partials/_step-nameAndNumbers.scss';
```
azefiel: This import should be nested inside the common control root selector to avoid the possibility of targeting unwanted elements (global pollution).

src/js/lib/monster.ui.js
```diff
@@ -1350,7 +1350,16 @@ define(function(require) {
...
+				var isString = _.isString(value),
+					isArray = _.isArray(value),
+					isStringValid = isString && monster.util.getFormatPhoneNumber(value).isValid,
+					isArrayValid = isArray
+						&& _
+							.chain(value)
+							.map(monster.util.getFormatPhoneNumber)
+							.every('isValid')
+							.value();
+				return this.optional(element) || isStringValid || isArrayValid;
```
azefiel: Nice, that's functional. Maybe we could improve this block by writing our check only once (right now it is written twice, for each type).

To do so, we could wrap `value` into an array no matter what and flatten it, such as:

```js
return this.optional(element) || _
	.chain([value])
	.flatten()
	.map(monster.util.getFormatPhoneNumber)
	.every('isValid')
	.value();
```

PR #521, opened 2019-12-02
MSTR-92: Upload CSV file and extract phone numbers

guillegr123:


src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -263,6 +268,175 @@ define(function(require) {
...
+
+					$numbersArea.val(function(i, text) {
+						var trimmedText = _.trim(text);
+
+						if (!(_.isEmpty(trimmedText) || _.endsWith(trimmedText, ','))) {
+							trimmedText += ', ';
+						}
+
+						return trimmedText + _.join(numbers, ', ');
+					}).focus();
```
azefiel: Not sure this logic belongs here as it is not really related to the file parsing.

What do you think of moving it to the event callback that triggered the file upload in the first place?

guillegr123: I agree with you

guillegr123:
No worries, thank you very much for your comments.

PR #524, opened 2019-12-06
MSTR-85: Lookup and validate phone number carriers

guillegr123:


src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -9,6 +9,12 @@ define(function(require) {
 
 		// Defines API requests not included in the SDK
 		requests: {
+			'phonebook.lookupNumbers': {
+				apiRoot: monster.config.api.phonebook,
+				url: 'lnp/lookup',
+				verb: 'POST',
+				generateError: false
```
masmerino13: Is this parameter needed?

guillegr123: In this case, yes. A custom message is shown if the request fails, to let the user know that the Phonebook service may be unavailable.

src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -505,6 +593,68 @@ define(function(require) {
...
+									.chain(numberCarriers)
+									.filter('portability')
+									.map('name')
+									.value();
+
+								return _.isEmpty(commonCarriers)
+									? portableCarriers
+									: _.intersection(commonCarriers, portableCarriers);
+							}, [])
+							.value();
```
azefiel: This piece of logic seems a bit convoluted as well; as I understand it, it returns an array containing carriers allowing porting that are common to all numbers.

To make this more clear, we could extract the filtering/mapping from the reduce:
```js
_.map(numbersData, function(numberData) {
    return _
        .chain(numberData.carriers)
        .filter('portability')
        .map('name')
        .value()
})
```
This would return an array of arrays, each containing the list of carriers allowing porting for a specific number. We could then spread that into `lodash#intersection` to retrieve the common carriers:
```js
_
.chain(numbersData)
.map(function(numberData) {
    return _
        .chain(numberData.carriers)
        .filter('portability')
        .map('name')
        .value()
})
.thru(_.spread(_.intersection))
.value()
```
Now this is somewhat easier to follow but still kind of hard to understand without context, so we could use a named function expression for `_.spread(_.intersection)` and give it a more meaningful name:
```js
var findCommonCarriers = _.spread(_.intersection);
_
.chain(numbersData)
.map(function(numberData) {
    return _
        .chain(numberData.carriers)
        .filter('portability')
        .map('name')
        .value()
})
.thru(findCommonCarriers)
.value()
```
As always, let me know if you have any questions for me regarding this!

src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -478,13 +484,109 @@ define(function(require) {
 		 * @param  {Function} callback  Callback to pass the step template to be rendered
 		 */
 		portWizardCarrierSelectionRender: function(args, callback) {
-			var self = this;
+			var self = this,
+				formattedNumbers = _.get(args.data, 'nameAndNumbers.numbersToPort.formattedNumbers'),
+				numbers = _.map(formattedNumbers, 'e164Number'),
+				initTemplateMultiple = function(dataTemplate) {
+					dataTemplate.errorType = 'noWinnerCarriers';
```
azefiel:
```suggestion
					dataTemplate.errorType = 'noWinningCarriers';
```

Also, you should use `merge#lodash` to create a new object rather than mutating the existing one.

src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -674,6 +838,40 @@ define(function(require) {
...
+				success: function(data) {
+					args.success(data.data);
+				},
+				error: function(data, status) {
+					var errorData = {
+						phonebookUnavailable: data.status === 0
+					};
+
+					_.has(args, 'error') && args.error(errorData);
+				}
```
azefiel: When you bypass the global error handler, you should always call it once your custom error handling is done to catch unexpected errors.

In this case, you might also want to add `500` to the list of status codes to handle.

src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -478,13 +484,95 @@ define(function(require) {
...
+					var numbersByLosingCarrier = carrierData.numbersByLosingCarrier,
+						winningCarriers = carrierData.winningCarriers,
+						losingCarriersCount = _.size(numbersByLosingCarrier),
+						isSingleLosingCarrier = losingCarriersCount === 1,
+						isSingleLosingCarrierUnknown = _
+							.chain(numbersByLosingCarrier)
+							.keys()
+							.head()
+							.isEqual('Unknown')
+							.value(),
```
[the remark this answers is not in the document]
guillegr123: Thanks, it was really convoluted

src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -505,6 +593,68 @@ define(function(require) {
 			};
 		},
 
+		/**
+		 * Get the losing and winning carriers for the phone numbers
+		 * @param  {Object} args
+		 * @param  {String} args.numbers  Phone numbers to look up
```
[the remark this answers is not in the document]
guillegr123: Thanks for the heads up! The type was wrong, the parameter is supposed to be an array of strings.

src/apps/common/submodules/portWizard/portWizard.js
guillegr123: That is nice! It will be a lot more readable that way.
Thanks also for the suggestion on the `lodash#spread` usage, I was wondering how to pass the array items as parameters for the `lodash#intersection` function and similar ones.

src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -674,6 +826,42 @@ define(function(require) {
...
+				},
+				error: function(data, error, globalHandler) {
+					var errorData = {
+						phonebookUnavailable: _.includes([0, 500], error.status)
+					};
+
+					_.has(args, 'error') && args.error(errorData);
+
+					globalHandler(data);
+				}
```
azefiel: Ah, my bad. I didn't realize your custom error callback was meant to handle every error.

In that case, you should make it mandatory, as it will always bypass `globalHandler`, something like:
```js
function(data, error) {
  args.error({
    isPhonebookUnavailable: _.includes([0, 500], error.status)
  });
}
```
I also took the opportunity to prepend `is` to the prop name, to denote the fact that it contains a boolean value, which later on makes is easier to use (`phonebookUnavailable` could contain the error payload for example).

guillegr123: My bad too, that's right, the error callback should be mandatory if it will handle every error :sweat_smile:

PR #525, opened 2019-12-07
KZOO-16: Fix port_authority property reading from port_request object

harenson:
Right now even if the account you are currently logged in is the port authority for a given port request the UI doesn't show the `action_required` nor the `private` options when you try to create a comment on that port request, this change fixes that.

### DIFF
--- a/src/apps/common/submodules/portListing/portListing.js
+++ b/src/apps/common/submodules/portListing/portListing.js
@@ -312,7 +312,7 @@ define(function(require) {
 							_.get(portRequest, 'numbers')
 						),
 						unactionableStatuses = ['canceled', 'completed'],
-						isAgent = self.portListingUtilIsAgent(_.get(portRequest, 'port_authority')),
+						isAgent = self.portListingUtilIsAgent(_.get(portRequest, '_read_only.port_authority')),
 						isUpdateable = !_.includes(unactionableStatuses, portRequest.port_state);
 
 					return _.merge({


src/apps/common/submodules/portListing/portListing.js
```diff
@@ -312,7 +312,11 @@ define(function(require) {
 							_.get(portRequest, 'numbers')
 						),
 						unactionableStatuses = ['canceled', 'completed'],
-						isAgent = self.portListingUtilIsAgent(_.get(portRequest, 'port_authority')),
+						isAgent = self.portListingUtilIsAgent(_.get(
+							portRequest,
+							'port_authority',
+							_.get(portRequest, '_read_only.port_authority'))
+						),
```
azefiel: Wouldn't it make sense to check the read only value directly then? Or is in not exposed at all times?

harenson: Right, it looks it is not including that `port_authority` key at the top of the json object anymore https://github.com/2600hz/kazoo/blob/master/core/kazoo_numbers/src/knm_port_request.erl#L100-L108

azefiel: In that case, could you update your change to simply:
```js
isAgent = self.portListingUtilIsAgent(_.get(portRequest, '_read_only.port_authority')),
```

harenson: Done, thank you for reviewing 👍

PR #527, opened 2019-12-09
MSTR-114: Add markdown editor

masmerino13:


src/js/lib/monster.ui.js
```diff
@@ -1465,6 +1466,28 @@ define(function(require) {
 			return form.valid();
 		},
 
+		/**
+		 * Convert the 'target' in to Markdown editor
+		 * @param target - mandatory jQuery object
+		 * @param [options] - optional object to overwrite the default configs
+		 */
```
azefiel: Our JSDoc method header provide at least both the [name and type](https://devdocs.io/jsdoc/tags-param#names-types-and-descriptions) of each parameter.

In this case, `target` is of type `jQuery` and `options` is of type `Object`.

src/js/lib/monster.ui.js
```diff
@@ -1465,6 +1466,28 @@ define(function(require) {
...
+		simplemde: function($target, options) {
+			if (!($target instanceof $)) {
+				throw TypeError('"$target" is not a jQuery object');
+			}
+
+			if (!_.isUndefined(options) && !_.isPlainObject(options)) {
+				throw TypeError('"options" is not a plain object');
+			}
+
+			new simplemde(_.merge({
```
azefiel: It's always a good idea for `monster.ui` helpers to return the instance created by the constructor to allow chaining.

masmerino13:
Good to know, let me revert that, thanks!

PR #540, opened 2019-12-24
MSTR-119: Allow link labels to wrap inside their container

azefiel:
Fixes link labels wrapping off their container on number entity selection.

image013


azefiel:
4.3 -> https://github.com/2600hz/monster-ui/pull/542

PR #546, opened 2020-01-08
MSTR-96: Implement ownership confirmation step

guillegr123:


src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -1021,18 +992,251 @@ define(function(require) {
...
+		 * @param  {Boolean} eventArgs.completeStep  Whether or not the current step will be
+		 *                                           completed
 		 * @returns  {Object}  Object that contains the updated step data, and if it is valid
 		 */
-		portWizardOwnershipConfirmationUtil: function($template, args) {
-			var self = this;
+		portWizardOwnershipConfirmationUtil: function($template, args, eventArgs) {
+			var self = this,
+				$form = $template.find('form'),
+				isValid = !eventArgs.completeStep || monster.ui.valid($form),
```
masmerino13: Is this working well?

since `isValid` will get the value of `!eventArgs.completeStep`, due to `eventArgs.completeStep` will be defined with `true or false`, which means this `OR` condition `(|| monster.ui.valid($form))` isn't going to happen.

Does that make sense?

guillegr123: Yes, it is working well. the form is validated (`monster.ui.valid($form)`) only when the step is going to be completed pressing the next button when it is the first time in the current step, or it is already completed and we are moving to any other step.

src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -1203,6 +1410,67 @@ define(function(require) {
...
+				btnText: i18n.labels.fileButton,
+				mimeTypes: fileRestrictions.mimeTypes,
+				maxSize: fileRestrictions.maxSize,
+				dataFormat: dataFormat,
+				success: args.success,
+				error: function(errorsList) {
+					var errorType = _
+							.chain(errorsList)
+							.pick('mimeTypes', 'size')
+							.omitBy(_.isEmpty)
```
masmerino13: :100:

PR #557, opened 2020-01-28
MSTR-127: Add notEqualTo custom validation rule

guillegr123:
This is a new custom rule for the jQuery validation plugin, to check that a form field’s value does not match one or more of other field’s values.


guillegr123:
4.3 -> #558

PR #563, opened 2020-01-30
MSTR-129: Enforce only one WS connection to be opened once logged in

azefiel:
Takes care of the following:

- Only attempt to open WS connection once a user is logged in
- Prevent multiple WS connections to be opened on reconnect

### DIFF
--- a/src/js/lib/monster.socket.js
+++ b/src/js/lib/monster.socket.js
@@ -40,11 +40,19 @@ define(function(require) {
 			}
 		},
 
-		initialize: function() {
+		/**
+		 * Attempts to open a WebSocket connection
+		 * @param  {Object} args
+		 * @param  {Function} [args.onError] Invoked when an error occurs
+		 */
+		initialize: function(args) {
 			var self = this,
 				socket;
 
+			self.isConnecting = true;
+
 			if (self.isConnected()) {
+				self.isConnecting = false;
 				self.log('Socket already active');
 				return;
 			}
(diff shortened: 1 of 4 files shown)


azefiel:
4.3 -> https://github.com/2600hz/monster-ui/pull/564

azefiel:
> Lgtm!
> 
> I think this might be outside of the scope of this PR, but it is just a heads up:
> Currently the second attempt to create a websocket connection may be performed while the first attempt is still in progress, because the `isConnecting` property is checked only before the first connection attempt, outside of the `connectAttempt` function. This sometimes result in two or more successful websocket connections, as a result of these multiple attempts. Could it be possible to execute the next connection attempt only until [the previous one fails](https://stackoverflow.com/questions/25779831/how-to-catch-websocket-connection-to-ws-xxxnn-failed-connection-closed-be)?

Good feedback, it indeed is pertinent to this PR. I believe my latest commit addresses your concerns.

Just for context, this is a hot fix to patch prod while the socket module is refactored to tick all of  [MSTR-129](https://2600hz-commercial.atlassian.net/browse/MSTR-129)'s boxes.

src/js/lib/monster.socket.js
```diff
@@ -90,45 +114,48 @@ define(function(require) {
...
 
-							self.log('Attempting to reconnect to WebSocket at ' + monster.util.toFriendlyDate(new Date()));
-							self.log('Next try in ' + timeoutDuration / 1000 + ' seconds');
-						} else {
-							timeoutDuration = startingTimeout;
+			if (!monster.util.isLoggedIn()) {
+				return self.log('Unable to connect to WebSocket while logged out');
+			}
+			if (self.isConnecting) {
+				return self.log('Already attempting to connect to WebSocket');
```
guillegr123: If the socket connection takes too long to fail, no more attempts will be done if this get executed in the meantime on a second attempt. Is this the expected behavior for the time being?

azefiel: It'd be nice not to introduce regressions, let me add an error callback to handle that.

PR #568, opened 2020-02-10
MSTR-100: Submit port request

guillegr123:


src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -1009,9 +1062,18 @@ define(function(require) {
...
 			var self = this,
+				requiredDocumentsCompleteList = self.portWizardGet('requiredDocumentsList'),
+				invoiceDocumentMetadata = _.find(requiredDocumentsCompleteList, { key: 'Invoice' }),
+				buildTemplateArgs = _
+					.chain(args)
+					.pick('data')
+					.merge({
+						invoiceDocumentMetadata: invoiceDocumentMetadata
+					})
+					.value(),
```
azefiel: On issue with this chaining is the fact that `buildTemplateArgs.data` holds a reference of `args.data` (shallow vs deep copy).

What you are doing:
```js
var buildTemplateArgs = {
    data: args.data,
    invoiceDocumentMetadata: invoiceDocumentMetadata
};
```

What you want to be doing:
```js
var buildTemplateArgs = {
    data: { ..args.data }
    invoiceDocumentMetadata: invoiceDocumentMetadata
};
```
I am using object spread here for conciseness but in lodash world this could be done by using your picked object as source rather than destination of your merge:
```js
var buildTemplateArgs = _.merge({
    invoiceDocumentMetadata: invoiceDocumentMetadata
}, _.pick(args, 'data'));
```

src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -1049,10 +1111,13 @@ define(function(require) {
...
 		/**
 		 * Get the template for Ownership Confirmation step (bill upload view)
-		 * @param  {Object} args  Wizard args
+		 * @param  {Object} args
+		 * @param  {Object} args.data  Wizard data
+		 * @param  {Object} args.invoiceDocumentMetadata Invoice document metadata
 		 */
 		portWizardOwnershipConfirmationBillUploadGetTemplate: function(args) {
 			var self = this,
+				invoiceDocumentMetadata = args.invoiceDocumentMetadata,
```
azefiel: Not sure why `invoiceDocumentMetadata` gets passed down two methods when `portWizardOwnershipConfirmationBillUploadBindEvents` is the only one that makes use of it.

src/apps/common/submodules/portWizard/portWizard.js
guillegr123: In this block I wanted to have a shalllow copy as a way to save some memory space, because the `args.data` object is only read and never modified in the `GetTemplate` functions below. Is it OK, or a deep copy should always be created in this kind of scenarios, to prevent any unintended side effects?

src/apps/common/submodules/portWizard/portWizard.js
guillegr123: It is used at `portWizardOwnershipConfirmationBillUploadBindEvents` and `portWizardOwnershipConfirmationAccountInfoBindEvents`. As the information is obtained from the submodule store, do you think it would be better to get the invoice document metadata whithin each method, instead of passing it all the way down from `portWizardOwnershipConfirmationRender`?

azefiel: > [...] do you think it would be better to get the invoice document metadata whithin each method, instead of passing it all the way down from `portWizardOwnershipConfirmationRender`?

From what I can tell, passing down `invoiceDocumentMetadata` adds complexity that could be avoided since that metadata is simply a map of static data.

I believe the problem stems from how `requiredDocuments.documents` is structured. Switching from an array to a map would make more sense and make it easier to retrieve the attachment name (you could use `lodash#get` or make a util for it to abstract the path).

If the order of `requiredDocuments.documents` matters, then it should be broken down into two different data structures: one array to hold the order of documents IDs and another to map document IDs to attachment name.

src/apps/common/submodules/portWizard/portWizard.js
azefiel: > Is it OK, or a deep copy should always be created in this kind of scenarios, to prevent any unintended side effects?

Keeping your data immutable (and, as you noted, prevent side effects) trumps memory complexity in almost any scenario; edge cases usually involve recursion, in which case specific optimization techniques can be implemented when your algorithm becomes a memory hog (memoization or switching to a loop statement).

> In this block I wanted to have a shalllow copy as a way to save some memory space, because the `args.data` object is only read and never modified in the `GetTemplate` functions below.

While this idea makes sense now, it is because this part of the codebase is fresh in your memory. Imagine having to come back to it in three months, or someone else modifying the structure of `buildTemplateArgs` downstream without realizing `args.data` will be modified as well.

PR #569, opened 2020-02-17
MSTR-101: Allow to save port request draft

guillegr123:


src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -2214,36 +2248,11 @@ define(function(require) {
...
-						email: {
-							send_to: _.size(notificationEmails) === 1
-								? _.head(notificationEmails)
-								: notificationEmails
-						}
 					}
-				});
-
+				}, billSection, transferDateSection, notificationsSection);
+			debugger;
```
azefiel:
```suggestion
```

src/apps/common/submodules/portWizard/portWizard.js
```diff
@@ -368,8 +368,9 @@ define(function(require) {
 				}),
 				title: i18n.title,
 				cancel: 'portWizardClose',
-				done: 'portWizardSubmit',
+				done: self.portWizardComplete,
```
azefiel: Curious why the method is referenced directly when `cancel` references the method's name?

guillegr123: I'm trying to update some methods to support references, while keeping the name support for backwards compatibility (i.e. `navigationWizard` instances already in use in CCPro and AppExchange). However I haven't updated the `cancel` method yet. I think I'll take the chance to do it now, because it is the last wizard global method that does not support references.

src/apps/common/submodules/portWizard/portWizard.js
guillegr123: :scream:

guillegr123: Just found that I left it in the penultimate commit, but then removed it on the last one: 
https://github.com/2600hz/monster-ui/pull/569/commits/2b48b9ea6e5d5fcc98b8a150d6faec6445a45d53

azefiel: Gotcha, I was reviewing commit per commit for context and must have missed that.

P.S.: penultimate means one before last

src/apps/common/submodules/portWizard/portWizard.js
azefiel: 👍  just be sure to take into account the context in which those methods will be called (`this`).

PR #574, opened 2020-02-25
MSTR-141: Print WebSocket related logs on dev environment only

azefiel:


azefiel:
4.3 -> https://github.com/2600hz/monster-ui/pull/577

PR #584, opened 2020-03-04
MSTR-146: Fix exceeded chars validation

masmerino13:


masmerino13:
Sure,

This fix the case where you have a paragraph with the exact number of allowed characters, for example:

summary = 100
allowedSize = 100

Actually we check the allowed caracters like this:
ref: ref: https://github.com/2600hz/monster-ui/blob/master/src/js/lib/monster.ui.js#L3873

`allowedChars = summary <= allowedSize ? allowedSize - summary : 0;`

This should return 0

And this condition is going to show that there is an error, saying that whe allowed chars was exceeded https://github.com/2600hz/monster-ui/blob/master/src/js/lib/monster.ui.js#L3877, but it's not correct, since the user have an exact number of allowed characters.

Does that make sense?

@azefiel

src/js/lib/monster.ui.js
```diff
@@ -3870,11 +3870,12 @@ define(function(require) {
...
 				var currentLenght = checkCurrentLength(),
-					allowedChars = currentLenght <= size ? size - currentLenght : 0;
+					allowedChars = currentLenght <= size ? size - currentLenght : -1,
+					allowedCharsText = allowedChars < 0 ? 0 : allowedChars;
 
-				allowedCharsLabel.text(allowedChars);
+				allowedCharsLabel.text(allowedCharsText);
 
-				if (allowedChars <= 0 || allowedChars > size) {
+				if (allowedChars < 0 || allowedChars > size) {
```
azefiel: Okay, so the problem is that we are using `allowedChars` for two different purposes:

- hold the remainder of chars available
- determine if current size exceeds max size.

To clarify the logic, we could use two different variables:

```js
const enforceMaxSizeFactory = maxSize => currentSize => ({
	remainder: Math.max(0, maxSize - currentSize),
	isGreaterThanMaxSize: currentSize > maxSize
});

const max = 100;
const enforceMaxSize = enforceMaxSizeFactory(max);

[99, 100, 101].forEach(
	curr => console.log({ curr, max, ...enforceMaxSize(curr) })
);

/*
{ curr: 99, max: 100, remainder: 1, isGreaterThanMaxSize: false }
{ curr: 100, max: 100, remainder: 0, isGreaterThanMaxSize: false }
{ curr: 101, max: 100, remainder: 0, isGreaterThanMaxSize: true }
*/
```
Which, transposed to our situation, would look like:
```js
checkLength = function(event) {
	var currentLength = checkCurrentLength(),
		isGreaterThanMaxSize = currentLength > size,
		allowedChars = Math.max(0, size - currentLength);

	allowedCharsLabel.text(allowedChars);

	if (isGreaterThanMaxSize) {
		// error handling
    }
```

masmerino13: Thank you @azefiel :100:

PR #588, opened 2020-03-06
MSTR-103: Allow to delete port request draft

guillegr123:


src/apps/common/submodules/navigationWizard/navigationWizard.js
```diff
@@ -120,7 +129,8 @@ define(function(require) {
 					{},
 					navigationWizardFlagsDefaults,
 					_.pick(args, 'askForConfirmationBeforeExit', 'currentStep', 'validateOnStepChange')
-				);
+				),
+				buttonProps = _.get(args, 'buttonProps', []);
 
 			if (!container) {
 				throw new Error('A container must be provided.');
```
masmerino13: It will nice to have this text in the i18n translation file

guillegr123: I've checked other source files, and it seems that the error messages do not have any translations, but are coded as is.

PR #589, opened 2020-03-06
Update phone number features resolver to support new path

azefiel:


azefiel:
4.3 -> https://github.com/2600hz/monster-ui/pull/590

src/js/lib/monster.util.js
```diff
@@ -1200,18 +1200,21 @@ define(function(require) {
...
 	function getNumberFeatures(number) {
 		if (!_.isPlainObject(number)) {
 			throw new TypeError('"number" is not an object');
 		}
-		var numberFeatures = _.get(number, 'features_available', []);
-		return _.isEmpty(numberFeatures)
-			? _.get(number, '_read_only.features_available', [])
-			: numberFeatures;
+		var pathToFeatures = _.find([
+			'_read_only.features.available',
```
guillegr123: Is this OK? The old path was `_read_only.features_available` (underscore instead of dot). Also, the order in which the paths are checked is inverted, because previously `features_available` was checked first.

azefiel: > Is this OK? The old path was `_read_only.features_available`

Yes, sorry if it wasn't clear but that's the whole point of this PR. The path was updated by https://github.com/2600hz/kazoo/pull/6291



> the order in which the paths are checked is inverted, because previously `features_available` was checked first.

That does not matter because:

- `features_available` is present when the number comes from `GET /phone_numbers`
- `_read_only.features.available` when it comes from `GET /phone_numbers/{PHONE_NUMBER}`

PR #602, opened 2020-03-30
Add account name to topbar

jon-braz:
- Added the current account name to the main topbar to make it easier to know which account you are logged into at a glance
- Added setAccount() to auth to update the account name in the DOM when it's changed (on login, masquerading)

Screen Shot 2020-03-30 at 11 27 45 AM


azefiel:
Hey @jon-braz, thanks for bringing up this issue. It is true that being able to see the account name you currently are logged in to at a glance would be nice.

After discussing this further with our design team, we've come up with the following mockup, which in our mind would help:
- save horizontal space in the topbar
- give better context into which account you are logged in to
- differentiate better when a user is masquerading vs just logged in
control-center-launcher

Please let me know if you'd be interested into reworking your PR to implement this new control center launcher. If not, no problem, we'll create an internal ticket for this work but it might not be worked on right away.

jon-braz:
@azefiel Sure, I've made the changes to match the new design. What were you thinking on hover? As is, I've built it so that only the account name gets $primary-color.


Screen Shot 2020-05-11 at 6 00 36 PM
Screen Shot 2020-05-11 at 6 00 46 PM

azefiel:
Nice, good job @jon-braz 

If you could update our base partial file, such as:

`src/css/partials/_base.scss`
```diff
$yellow-sea: #F8A506;
+ $ecstasy: #FD720F;
$red-orange: #FF3D24;

...

$primary-color: $dodger-blue;
+ $secondary-color: $ecstasy;
$danger-color: $red-orange;
```
You can then use `$secondary-color` as the masquerading orange color.

Regarding the on hover action, it's fine and matches the behavior of other topbar actions (we'll have a refresh of that once this PR gets merged).

jon-braz:
Sounds good, I've updated the colors

Screen Shot 2020-05-13 at 9 26 04 AM

azefiel:
Perfect 🎉 🎉 🎉 

Thanks again for your contribution!

(This feature should be packaged into the next release)

jon-braz:
Thanks, happy to help :)

PR #622, opened 2020-04-20
MPRO-7: For Polycom Models, Prefix VVX in front of all VVX model names in Monster UI

jrmende5:
For Polycom Models, Prefix VVX in front of all VVX model names in Monster UI

The family information is already available on the payload, it just need to be added to the string displayed to avoid confusions.

### DIFF
--- a/src/apps/common/submodules/chooseModel/chooseModel.js
+++ b/src/apps/common/submodules/chooseModel/chooseModel.js
@@ -41,6 +41,7 @@ define(function(require) {
 				families,
 				models,
 				prefix = ['vvx'];
+				
 			_.each(data, function(brand, brandKey) {
 				families = [];
 
@@ -48,17 +49,10 @@ define(function(require) {
 					models = [];
 
 					_.each(family.models, function(model, modelKey) {
-						if ( prefix.includes(family.name) ) {
-							models.push({
-								id: modelKey.toLowerCase(),
-								name: _.toUpper(family.name) + " " + model.name
-							});
-						} else {
-							models.push({
-								id: modelKey.toLowerCase(),
-								name: model.name
-							});
-						}
+						models.push({
+							id: modelKey.toLowerCase(),
+							name: prefix.includes(family.name) ? _.toUpper(family.name) + " " + model.name : model.name
+						});
 					});
 
 					families.push({
--- a/src/apps/common/submodules/chooseModel/chooseModel.js
+++ b/src/apps/common/submodules/chooseModel/chooseModel.js
@@ -39,8 +39,8 @@ define(function(require) {
 					brands: []
 				},
 				families,
-				models;
-
+				models,
+				prefix = ['vvx'];
 			_.each(data, function(brand, brandKey) {
 				families = [];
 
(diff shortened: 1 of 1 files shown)


src/apps/common/submodules/chooseModel/chooseModel.js
```diff
@@ -39,19 +39,26 @@ define(function(require) {
...
 
 				_.each(brand.families, function(family, familyKey) {
 					models = [];
 
 					_.each(family.models, function(model, modelKey) {
-						models.push({
-							id: modelKey.toLowerCase(),
-							name: model.name
-						});
+						if ( prefix.includes(family.name) ) {
```
masmerino13, line 51: Since there is no a big diference between the if/else statement I think this can be a little short, like this
```
models.push({
    id: modelKey.toLowerCase(),
    name: prefix.includes(family.name) ? _.toUpper(family.name) + " " + model.name : model.name
});
```

by this way, the if/else statement are not needed

src/apps/common/submodules/chooseModel/chooseModel.js
```diff
@@ -39,19 +39,26 @@ define(function(require) {
 					brands: []
 				},
 				families,
-				models;
-
+				models,
+				prefix = ['vvx'];
```
masmerino13, line 43: I think the line break is needed just to indicate that there is a new block of code

src/apps/common/submodules/chooseModel/chooseModel.js
jrmende5: changed

src/apps/common/submodules/chooseModel/chooseModel.js
jrmende5: changed

azefiel:
Thanks for the PR @jrmende5 and apologies for the delayed resolution.

@masmerino13 could you please review this again? Since this PR was updating imperative code, I took the opportunity to refactor a bit.

src/apps/common/submodules/chooseModel/chooseModel.js
```diff
@@ -35,40 +47,39 @@ define(function(require) {
...
+						id: _.lowerCase(brandKey),
+						name: brand.name,
+						families: _.map(brand.families, function(family, familyKey) {
+							return {
+								id: _.toLower(familyKey),
+								name: family.name,
+								models: _.map(family.models, function(model, modelKey) {
+									return {
+										id: _.toLower(modelKey),
+										name: resolveModelName(family.name, model.name)
```
masmerino13, line 75: Thanks for that, Looks great to me.

I'm curious about something.

How this should works for prefixes that are not included in `familyNamesAsPrefix` array?
Looking at `resolveModelName` this method looks for prefixes to return the model name, but if the model family isn't in the list, what this method will return, since this is filtering by `familyName` I assume an empty value `string`?

Please, let me know if I'm missing something :)

@jrmende5 @azefiel

azefiel: `lodash#filter` returns an array not matter what, even when no item pass the test.

So when the model's family is not included in `familyNamesAsPrefix`, then an empty array will be concatenated to the array containing the model name, which is joined and gives a final string representing the model's name to display to the end-user.

masmerino13: Ah, that makes more sense now, thanks for clarify.

PR #624, opened 2020-04-24
MSTR-162: Add ability to declare app links to be listed alongside apps

azefiel:


src/apps/apploader/app.js
```diff
@@ -140,21 +161,22 @@ define(function(require) {
...
-					parent.find('.app-description')
-							.find('h4')
-							.text(app.label)
-							.addBack()
-							.find('p')
-							.text(app.description);
+					var app = _.find(appList, { id: id }),
+						isLink = _.includes(_.keys(monster.config.whitelabel.appLinks), id);
+
+					parent.find('.app-description h4')[isLink ? 'addClass' : 'removeClass']('is-link');
```
guillegr123, line 167: I think it would be better to save the reference for the element `.app-description` as a variable at `bindEvents`, to reduce the search scope each time an app icon is hovered.

src/apps/apploader/app.js
```diff
@@ -80,58 +80,79 @@ define(function(require) {
...
-						});
-					}
-					args.callback && args.callback();
-				});
-			} else {
-				self.show({
-					callback: args.callback
-				});
-			}
+					callback(null, appList);
```
guillegr123, line 132: `appList` is not being used in the final callback for `monster#waterfall`. I think it can be safely removed from these callback arguments.

src/apps/apploader/app.js
azefiel: Good call, fixed!

PR #643, opened 2020-05-08
[4.3] MSTR-166: Set app selector dialog height relative to viewport

guillegr123:


azefiel:
Closing as the new `latest` target is now `5.0`.

PR #648, opened 2020-05-14
Refactor topbar style

azefiel:


azefiel:
Good catch, I updated links' icons height without realizing the masquerading one was larger.

Also, it seems like there is a weird size overflow behavior when the viewport's width is between 1200px and 1279px so I'll have to investigate that in more detail before merging.

src/apps/core/app.js
```diff
@@ -148,92 +148,43 @@ define(function(require) {
...
+						.map(function(app) {
+							return _
+								.chain({
+									icon: monster.util.getAppIconPath(app)
+								})
+								.merge(app)
+								.thru(monster.ui.formatIconApp)
+								.value();
+						})
+						.find({ name: name })
```
guillegr123, line 178: It seems that there is a double filter step here. I think this can be simplified to:
```suggestion
						.find({ name: name })
						.thru(function(app) {
							return _
								.chain({
									icon: monster.util.getAppIconPath(app)
								})
								.merge(app)
								.thru(monster.ui.formatIconApp)
								.value();
						})
```

azefiel: Arrays usually are better structures to work with in streams to gracefully handle unexpected errors.

When applying `find` right away and in case `name` does not match any items in the array resulting from `concat`,  the app would crash (because `lodash#find` returns `undefined` when no element matches its predicate).

We could use a ternary to branch between two values and avoid this bug, or use `filter` and `find`, which ultimately achieve the same result.

PR #652, opened 2020-05-20
Check for other open dialogs before allowing body to scroll

azefiel:


src/js/lib/monster.ui.js
```diff
@@ -821,14 +821,27 @@ define(function(require) {
...
 				// Event handlers
 				close: function() {
+					var hasOtherDialogsOpen = _
+						.chain($body.find('.ui-dialog .ui-dialog-content'))
+						.map(function(element) {
+							return $(element).dialog('isOpen');
+						})
+						.some(function(isOpen) {
+							return isOpen === true;
+						})
```
guillegr123, line 831: Just to be sure, the extra function is because the element may not be a dialog, and `isOpen` may not return a boolean? Because if the previous map already return booleans, `some` could be invoked without any parameter:
```suggestion
						.some()
```

azefiel: Coming back to this made me realize both operations can be solidified into a single `some` predicate, thanks for the pointer!

PR #656, opened 2020-05-28
Support nested flat ternary expression

azefiel:


src/.eslintrc.yml
```diff
@@ -20,7 +20,12 @@ globals:
 rules:
   standard/no-callback-literal: off
   no-tabs: off
-  indent: [ "error", "tab", { "SwitchCase": 1, "MemberExpression": "off" }]
+  indent:
+    - error
+    - tab
+    - SwitchCase: 1
+    - MemberExpression: off
+    - flatTernaryExpressions: true
```
guillegr123, line 28: The three properties should be part of the same item in the array:
```suggestion
    - SwitchCase: 1
      MemberExpression: off
      flatTernaryExpressions: true
```
If not, the next error is shown when running the lint task:
```
Error in plugin "gulp-eslint"
Message:
    /.../monster-ui/src/.eslintrc.yml:
        Configuration for rule "indent" is invalid:
        Value ["tab",{"SwitchCase":1},{"MemberExpression":"off"},{"flatTernaryExpressions":true}] should NOT have more than 2 items.
```

PR #662, opened 2020-06-01
MSTR-171: Request URL parameters should be encoded by default

pcandia:


src/js/lib/jquery.kazoosdk.js
```diff
@@ -621,7 +621,7 @@
 							if (methodInfo.verb.toLowerCase() === 'post' && k === ids.length - 1 && !(v in methodSettings)) {
 								requestSettings.data[v] = requestSettings.data.data.id;
 							} else {
-								requestSettings.data[v] = staticValues[v];
+								requestSettings.data[v] = encodeURIComponent(staticValues[v]);
```
azefiel, line 624: While this does encode the component, we end up storing the result, which might be problematic if more operations get performed on it down the road (encoded values only makes sense in the context of an URL).

Moving the encoding operation directly to where we [pattern match](https://github.com/2600hz/monster-ui/blob/358460fbb69ac92db6ab84065a05f3b85670cb41/src/js/lib/jquery.kazoosdk.js#L711) request parameters (similar to what you did for filters) would be a better solution.

PR #667, opened 2020-06-02
MSTR-172: Specify account for app selector list

guillegr123:
4.3 -> #668

### DIFF
--- a/src/apps/apploader/app.js
+++ b/src/apps/apploader/app.js
@@ -491,7 +491,7 @@ define(function(require) {
 
 			monster.waterfall([
 				// Get all apps
-				function(waterfallCallback) {
+				function getAllApps(waterfallCallback) {
 					var allAps = monster.appsStore;
 
 					if (!forceFetch && !_.isEmpty(allAps)) {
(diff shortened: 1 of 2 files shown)


src/apps/apploader/app.js
```diff
@@ -534,25 +534,17 @@ define(function(require) {
...
-							return false;
+							var appAllowedUsers = _.get(app, 'allowed_users'),
+								areAllUsersAllowed = appAllowedUsers === 'all',
+								isAdminUserAllowed = appAllowedUsers === 'admins' && currentUser.priv_level === 'admin',
+								isCurrentUserAllowed = appAllowedUsers === 'specific'
+									&& _
+										.chain(app)
+										.get('users', [])
+										.some({ id: currentUser.id })
+										.value();
```
azefiel, line 545: Why not store this result in a variable as well, or use the same `lodash#include`  check than before? It might make it easier to read.

Also, in this case using `lodash#find` vs `lodash#some`  combined with `lodash#isUndefined` would make more sense (although it has the same effect in this case).

guillegr123: Sorry, I missed this comments a couple of days. Sounds good. I went with the first approach, it will make it easy to read.

PR #669, opened 2020-06-02
MSTR-169: Add JSONEditor library

FerjoCode:


src/js/lib/monster.ui.js
```diff
@@ -1525,6 +1526,37 @@ define(function(require) {
...
+			$target
+				.addClass(customClass ? customClass : 'monster-jsoneditor');
+
+			var container = $target[0],
+				jsonObject = json || {},
+				formattedOptions = _.merge({}, {
+					mode: 'code',
+					enableSort: false,
+					enableTransform: false,
+					search: true
```
azefiel, line 1551: Could you explain why those should be set by default for every instance? (I'm curious and haven't read the editor's doc so just wondering).

src/css/lib/custom/jsoneditor-add.scss
```diff
@@ -0,0 +1,6 @@
+@import "../../partials/base";
+
+.monster-jsoneditor {
+	width: 600px;
+	height: 70vh;
+}
```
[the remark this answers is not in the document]
FerjoCode: Yeah you're right, I had added these styles since set an initial size to the editor's container is mandatory, but I can just point it out this in the docs and the user can handle it

src/js/lib/monster.ui.js
FerjoCode: True that. I thought It would keep the editor as simple as possible, but I can rid off these options and it don't mean any significant impact in the editor

FerjoCode:
I have also add the posibility to retrieve the editor's instance from the DOM and to make it easier I have also created another UI helper to do that

docs/monster-ui/jsoneditor().md
```diff
@@ -0,0 +1,68 @@
...
+## Syntax
+```javascript
+monster.ui.jsoneditor(target[,options])
+```
+
+### Parameters
+
+Key | Description | Type | Default | Required
+:-: | --- | :-: | :-: | :-:
+`target` | A Jquery object which is automatically contain the JSON Editor | `jQuery` | |`true`
```
azefiel, line 17: The "Type" column already provides the type of this parameter, no need to repeat yourself.
```suggestion
`target` | Container for the JSON Editor widget. | `jQuery` | |`true`
```

docs/monster-ui/jsoneditor().md
```diff
@@ -0,0 +1,68 @@
...
+`options` | Let you override default editor options that can be found on the [JSONEditor](https://github.com/josdejong/jsoneditor/blob/master/docs/api.md) page | ([#options](#options)) | |`false`
```
azefiel, line 18:
```suggestion
`options` | Let you override default editor options that can be found on the [JSONEditor](https://github.com/josdejong/jsoneditor/blob/master/docs/api.md) page | `Object`([#options](#options)) | |`false`
```

docs/monster-ui/jsoneditor().md
```diff
@@ -0,0 +1,68 @@
...
+
+### `options`
+Here is a list of the customization settings specific to Monster UI.
```
azefiel, line 21:
```suggestion
List of helper specific customizable settings (in opposition to the ones exposed by the JSON Editor library).
```

docs/monster-ui/jsoneditor().md
```diff
@@ -0,0 +1,68 @@
...
+
+Key | Description | Type | Default | Required
+:-: | --- | :-: | :-: | :-:
+`json` | Default JSON to set in the editor | `Object` | | `false`
+
+### Return
+An instance of the JSONEditor
+
+## Description
+The `monster.ui.jsoneditor()` method renders a powerfull jsoneditor into the specified `target` which is completely customizable passing the `options` parameter. This method returns the `JSONEditor` instance so you have access to its properties and methods. Please see [JSONEditor](https://github.com/josdejong/jsoneditor) for reference. To get the JSONEditor instance from the DOM please see [getJsoneditor()](./getJsoneditor().md)
```
azefiel, line 31:
```suggestion
The `monster.ui.jsoneditor()` method renders a customizable JSON editor into the specified `target`. This method returns the `JSONEditor` instance so you have access its properties and methods. Please see [JSONEditor](https://github.com/josdejong/jsoneditor) for reference. To get the JSONEditor instance from the DOM please see [getJsoneditor()](./getJsoneditor().md)
```

docs/monster-ui/jsoneditor().md
```diff
@@ -0,0 +1,68 @@
...
+
+To initialize the JSONEditor, you just need to pass the `target` parameter:
+```html
+<!--You must have to set the size of the container-->
```
azefiel, line 35:
```suggestion
<!--You need to set the container's size -->
```

docs/monster-ui/jsoneditor().md
```diff
@@ -0,0 +1,68 @@
...
+<div id="jsoneditor" style="width: 400px; height: 400px;"></div>
+```
+
+```javascript
+var $target = $('#jsoneditor');
+
+var jsoneditor = monster.ui.jsoneditor($target)
```
azefiel, line 42:
```suggestion
var jsoneditor = monster.ui.jsoneditor($target);
```

docs/monster-ui/getJsoneditor().md
```diff
@@ -0,0 +1,29 @@
+title: getJsoneditor()
+
+# monster.ui.getJsoneditor()
+The `monster.ui.getJsoneditor()` method gets a JSONEditor instance from the DOM with the editor's container
```
azefiel, line 4: No need for a tldr when the helper is this simple.

Also, please don't forget to end your sentences with periods.
```suggestion
```

docs/monster-ui/getJsoneditor().md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+### Parameters
+
+Key | Description | Type | Default | Required
+:-: | --- | :-: | :-: | :-:
+`target` | A Jquery object which contains the JSON Editor | `jQuery` | |`true`
+
+### Return
+An instance of the JSONEditor or null if the editor's container does not exist
```
azefiel, line 18:
```suggestion
An instance of the JSONEditor or `null` if the editor's container does not exist
```

docs/monster-ui/getJsoneditor().md
```diff
@@ -0,0 +1,29 @@
...
+
+## Description
+The `monster.ui.getJsoneditor()` method provides us access to the editor's instance from the DOM using the `target` parameter, which is the [JSONEditor](./jsoneditor().md) container. This method returns the `JSONEditor` instance.
```
azefiel, line 21:
```suggestion
The `monster.ui.getJsoneditor()` method provides access to the editor's instance from the DOM using the `target` parameter, which is the [JSONEditor](./jsoneditor().md) container. This method returns the `JSONEditor` instance.
```

PR #687, opened 2020-06-15
MSTR-177: Show app description tooltip if overflown

guillegr123:
4.3 -> #688

image


azefiel:
This has become stale and needs further input from design, closing for now.

PR #689, opened 2020-06-16
MSTR-179: Allow mediaSelect to support optional Text-to-Speech media tab

masmerino13:


src/apps/common/i18n/en-US.json
```diff
@@ -648,7 +648,12 @@
 		"silence": "Silence",
 		"shoutcastURLWarning": "WARNING: If the stream selected here attempts to be played, and is unavailable, the call may fail completely. Please make sure the stream you use is available 24/7.",
 		"useExistingMedia": "Use Existing Media",
-		"uploadNewMedia": "Upload New Media"
+		"uploadNewMedia": "Upload New Media",
+		"textToSpeech": {
+			"tab": "Text-to-Speech",
+			"info": "Customize your greeting message using Text-to-Speech. Type your desired greeting below. Once your changes are saved, anyone who calls this number will be greeted with your custom message.",
```
azefiel, line 654: We might want to insert a variable in place of `"number"` in this sentence, as TTS could be used for different entities (number, voicemail, conference etc.).

The value of this variable could either be set by the invocator directly (e.g. they would set the i18n string to be displayed) or, we could keep a list of entities locally. Which do you thing would make more sense?

```suggestion
			"info": "Customize your greeting message using Text-to-Speech. Type your desired greeting below. Once your changes are saved, anyone who calls this {{entity}} will be greeted with your custom message.",
```

src/apps/common/submodules/mediaSelect/views/tabs-layout.html
```diff
@@ -12,6 +12,13 @@
 						{{ i18n.mediaSelect.uploadNewMedia }}
 					</a>
 				</li>
+				{{#if enableTextspeechTab}}
+				<li class="navbar-menu-item">
+					<a href="javascript:void(0);" class="navbar-menu-item-link" data-tab="text_to_speech">
+						{{ i18n.mediaSelect.textToSpeech.tab }}
+					</a>
+				</li>
+				{{/if}}
```
guillegr123, line 21: It would be good to add some indentation to the `if` block, to improve readability and follow documents formatting.
```suggestion
			{{#if enableTextspeechTab}}
				<li class="navbar-menu-item">
					<a href="javascript:void(0);" class="navbar-menu-item-link" data-tab="text_to_speech">
						{{ i18n.mediaSelect.textToSpeech.tab }}
					</a>
				</li>
			{{/if}}
```

src/apps/common/submodules/mediaSelect/mediaSelect.css
```diff
@@ -78,3 +78,19 @@
...
+
+.monster-media-selection.monster-media-tabs-skin .monster-tab-content-tts .chars-remaining-counter {
+	font-size: 12px;
+	color: #64646c
+}
+
+.monster-media-selection.monster-media-tabs-skin .monster-tab-content-tts .chars-remaining-counter span {
+	font-weight: bold;
+	margin-right: 4px;
+}
```
azefiel, line 96: Please transition this sheet to scss so you can use nested rules and existing colors variables.

If you have time and since the amount of rules is pretty small, you can even refactor the whole sheet (as long as it's in its own specific "refactor" commit that it does not contain any other changes).

src/apps/common/submodules/mediaSelect/mediaSelect.js
```diff
@@ -113,32 +120,44 @@ define(function(require) {
 			} else if (skin === 'tabs') {
 				template = $(self.getTemplate({
 					name: 'tabs-layout',
-					data: formattedData,
+					data: _.merge(formattedData, {
```
guillegr123, line 123: It is better to merge values into a new object, in order to not to modify the original one that has been passed as a parameter, to avoid any unexpected side effects in the callers (unless it is the intention).
```suggestion
					data: _.merge({}, formattedData, {
```

src/apps/common/submodules/mediaSelect/mediaSelect.js
```diff
@@ -113,32 +120,44 @@ define(function(require) {
...
 
 				if (val === 'shoutcast') {
-					response = template.find('.shoutcast-div input').val();
+					callback(template.find('.shoutcast-div input').val());
+				} else if (ttsTab.length) {
+					args.tts.text = ttsTab.find('.custom-greeting-text').val();
+
+					self.mediaSelectUpdateTTSMedia(args, function(id) {
+						callback(id);
+					});
```
guillegr123, line 154: The callback function could be sent as a parameter as is:
```suggestion
					self.mediaSelectUpdateTTSMedia(args, callback);
```

src/apps/common/submodules/mediaSelect/mediaSelect.js
```diff
@@ -33,8 +38,8 @@ define(function(require) {
 				.append(template);
 
 			callback && callback({
-				getValue: function() {
-					return self.mediaSelectGetValue(template, args);
+				getValue: function(callback) {
+					self.mediaSelectGetValue(template, args, callback);
 				}
```
guillegr123, line 43: This could be shortened with lodash's help:
```suggestion
				getValue: _.partial(self.mediaSelectGetValue, template, args),
```

src/apps/common/submodules/mediaSelect/mediaSelect.js
```diff
@@ -294,6 +313,46 @@ define(function(require) {
 			monster.ui.fancyTabs(template.find('.monster-tab-wrapper'));
 
 			self.mediaSelectBindCommon(template, mediaToUpload);
+		},
+
+		mediaSelectUpdateTTSMedia: function(args, callback) {
+			var self = this,
+				greetingMedia = {
+					description: '<Text to Speech>',
```
guillegr123, line 321: If this is displayed to the user, it should be part of the `i18n` files.

src/apps/common/submodules/mediaSelect/mediaSelect.js
```diff
@@ -113,32 +120,44 @@ define(function(require) {
...
+					self.mediaSelectUpdateTTSMedia(args, function(id) {
+						callback(id);
+					});
 				} else {
-					response = val;
+					callback(val);
 				}
 			} else {
-				response = 'invalid_template';
+				callback('invalid_template');
```
azefiel, line 159: The problem with switching from sync to async is that it's not backward compatible.

It would make more sense to use a return statement when no `callback` parameter is provided to keep the existing method's signature valid. If a callback is provided, then we can use it to return the media id once created.

src/apps/common/submodules/mediaSelect/mediaSelect.js
```diff
@@ -20,6 +20,11 @@ define(function(require) {
 		 * @param {Array} args.options
 		 * @param {Array} args.mimeTypes - Allowed file formats to be uploaded if `dragableUpload` is true
 		 * @param {Boolean} [args.dragableUpload]
+		 * @param {Object} [args.tts] - Enable text to speech tab if present
+		 * @param {String} [args.tts.id] - Current media id, used when the media is updated
+		 * @param {String} [args.tts.text] - Current tts text
+		 * @param {String} [args.tts.name]
+		 * @param {String} [args.tts.type]
```
azefiel, line 27: Although not documented (😞) it looks like `args` already has a prop that contains the selected option (`selectedOption`), which we could use to hold the media object.

Then `tts` would be a simple boolean (could be renamed to `withTts` or something alike to denote the boolean value) to determine if the tab should be shown.

src/apps/common/submodules/mediaSelect/mediaSelect.js
masmerino13: No, it isn't

azefiel: Yeah, that [prop](https://docs.2600hz.com/supported/applications/crossbar/doc/media/#schema) is usually used to store the original file name, which doesn't really apply to TTS so we use a placeholder instead.

src/apps/common/i18n/en-US.json
masmerino13: hey @azefiel , I think using a i18n string will be the best way, just following the way that we already handle strings in Monster UI.

azefiel:
Converting back to draft as more work is required before MVP is achieved.

masmerino13:
Thanks @azefiel !

masmerino13:
Thanks @azefiel , I have fixed the bug

masmerino13:
Hi @azefiel ,

Thanks the comments, I have fixed the issues of points 1, 2 and 3. About the 4th I think this is because for TTS is needed to send the `tts` prop, you can see that in the method jsdocs.

So, the control usage should be something like this:

```
monster.pub('common.mediaSelect.render', {
    container: $('.app-content-wrapper'),
    skin: 'tabs',
    enableTTS: true,
    tts: {
        name: 'MainConferenceGreeting',
        type: 'mainConfGreeting',
        entity: '<entityHere>'
     }
    callback: function(control) {
    	control.getValue(console.log);
    }
});
```

Just let me know if you have any comments.

Thanks for your help!

azefiel:
Please test your changes live @masmerino13 , still getting an error when invoking with non-existing media ID, as well as a 400 when trying to create a new TTS media.

masmerino13:
Hi @azefiel ,

yes test this in the conference app, could you show me how are you generating those errors?

Thanks for your help!

azefiel:
Sure, here are the errros:
1. TTS upload error
mediaSelector-tts-upload-error
2. Invalid media error
mediaSelector-invalid-media-error

masmerino13:
Thanks @azefiel 

1.
Since to create a TTS media is required a type and name, when `enableTTS: true` you shoudl include the `tts` object which include those values, like so:
```
monster.pub('common.mediaSelect.render', {
    container: $('.app-content-wrapper'),
    skin: 'tabs',
    enableTTS: true,
    tts: {
        name: 'MainConferenceGreeting',
        type: 'mainConfGreeting',
        entity: '<entityHere>'
     }
    callback: function(control) {
    	control.getValue(console.log);
    }
});
```

2.
I will take a look on that, thanks

azefiel:
Okay, makes sense.

How about guarding against such an edge case though? A dev shouldn't be able to generate such an error when using the common control, even if this particular case is explained in the documentation.

masmerino13:
The name and type are required from the API so, those should be included, in that case, the other option is to have default values for those, but not sure if that will be correct, since probably a dev will assume that everithing is ok, but he will be storing wrong medias.

The other option is to show an alert error telling that this field is required.

I'm open to suggestions :)

azefiel:
I'd go with default values so the common control doesn't have to handle more errors cases.

Seems like `type` isn't mandatory so no need to add it when not specified by the invocator. As for `name`, we could add a default value similar to `description` when not specified. And `entity` could refer to `Entity` when not present.

What do you think?

masmerino13:
Yes, sounds good to me, what should be the default value? `<Text to Speech>`? about the entity value, this one isn't required, this is displayed in the text above the tts textarea.

azefiel:
`<Text to Speech>` sounds good.

What would happen if `entity` isn't specified? Isn't it used [here](https://github.com/2600hz/monster-ui/blob/3b1d70b55e7be6887e7330c1e2f68fdaa698d9d1/src/apps/common/i18n/en-US.json#L654)?

masmerino13:
Yes, it is, what say is this should looks like *anyone who calls this will be greeted with your custom message.* haha, but yeah,what i said that isn't required, but for sure, we should include a deafult value

masmerino13:
Thanks @azefiel , I did those changes

PR #694, opened 2020-06-24
add carrier api url

dariomalfatti-centropaghe:
I'm creating a new app to handle what was called "carrier" in the old ui (kazoo-ui api v1). I tested the API in the v2 sdk and they works as a charm but to use them with the kazoosdk I need them to be defined into the library. This tweak is just the api definitions.


azefiel:
Have you tried using the [`/resources` API](https://docs.2600hz.com/supported/applications/crossbar/doc/resources/)?

It should be a drop-in replacement for `/local_resources`.

dariomalfatti-centropaghe:
Oh yes, sorry, I miss them. They works perfectly. Thanks.

PR #708, opened 2020-07-23
[4.3] MSTR-179: Allow mediaSelect to support optional Text-to-Speech media tab

masmerino13:


azefiel:
Curious why this needs to be in `4.3`?

masmerino13:
I'm not sure, too ;), should not be here? I have created the other one for 5.0 as well

azefiel:
Since `5.0` is the new `latest`, `4.3` should only be targeted for bug fixes.

Closing.

PR #710, opened 2020-07-23
MSTR-181: Cast input to string before formatting

azefiel:
- Functionize method
- Streamline formatting logic
- Cast input to string


src/js/lib/monster.util.js
```diff
@@ -976,6 +965,26 @@ define(function(require) {
...
+	 * @return {String} Formatted string representation of the value.
+	 */
+	function formatVariableToDisplay(variable) {
+		return _
+			.chain(variable)
+			.toString()
+			.replace(/_/g, ' ')
+			.replace(/\w\S*/g, function(txt) {
+				return txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase();
+			})
```
guillegr123, line 983: Could `lodash#capitalize` be used instead?
```suggestion
			.replace(/\w\S*/g, _.capitalize)
```

azefiel: Sure, that works too!

PR #717, opened 2020-07-28
Unlock main thread on app load

azefiel:
The goal of this PR is to remove requests from the main thread wherever possible to prevent browser locking.

The first three commits take care of that, as well as clean up the logic by using waterfall control flow.

The last two commits profit from migrating those request to asynchronous handling and expose potential app loading errors to the original invokator (`monster.apps#load`).

Note: synchronous requests are still made when [retrieving templates from the server](https://github.com/2600hz/monster-ui/blob/master/src/js/templates.js#L30) but that would be complicated to refactor since `app#getTemplate` is synchronous as well.


azefiel:
The [5.0 based branch](https://github.com/2600hz/monster-ui/pull/718) needs review as well @guillegr123

PR #718, opened 2020-07-28
[5.0] Unlock main thread on app load

azefiel:


azefiel:
Since this PR will only moderatly improve performance and is not fixing a bug, I don't believe it should be backported @guillegr123

PR #735, opened 2020-08-17
MSTR-187: Enforce valid media ID conditionally

masmerino13:


masmerino13:
What you mean is that the validation message should be showed below the dropdown, right?

azefiel:
@masmerino13 yes, a rules should be added for that field and validation should take place using our form validator.

masmerino13:
thanks @azefiel , I have added the form validation

PR #762, opened 2020-09-19
Show control center when at least one tab is enabled

azefiel:


pcandia (approved):
looks good, no comments to add.

PR #785, opened 2020-10-20
MSTR-197: Abstract appstore metadata management into methods

azefiel:
The goal of this PR is to abstract away app store data management operations so that app metadata access is normalized. The purpose of this change is to allow for app metadata branding on top of internationalization resolution.

### `monster.util#getAppStoreMetadata`
Returns metadata for an appstore app formatted by resolving its icon path and i18n entry. It is this formatting logic that will be augmented later to resolve i18n based on branding configuration.
App metadata is retrieved from the current appstore, which corresponds to the current account. To perform the same operation for a different list of apps, we can use `monster.util#getAppMetadata` by providing it a custom list.

### `monster.util#listAppStoreMetadata`
Returns formatted apps metadata filtered for a specific scope (`all|account|user`). Similarly to `monster.util#getAppStoreMetadata`, the apps metadata it retrieved from the current appstore. When the `user` scope is used, apps will be filtered against the current user.

To perform the same listing operation on a specific list of apps, `monster.util.listAppsMetadata` can be used. To perform the same listing operation for a specific user, we can use `monster.util#listAppsMetadata` and filter the results with `monster.util#isUserPermittedApp` by providing it a user to filter against.

### `monster.util#getCurrentUserDefaultApp`
Resolves the default app for current user by checking the `appList`'s document prop against the appstore. When no valid app is found, get the first one accessible (if any) for the user from the appstore.

### `apploader#getAppList`
Was updated to use the new utils when fetching apps metadata for the current appstore. The `forceFetch` parameter was replaced by checking the presence of `accountId`. Only when `accountId` is different than the original account ID (for which the locally stored appstore corresponds to) does an API request is made to fetch the appstore.

It is also possible to provide a `user` parameter to filter the listing against. Note that when listing the appstore apps for the current appstore, when `user` is not provided but `scope='user'`, the listing will be filtered against the current user.

### `appstore#loadData`/`appstore#showAppPopup`
Use current appstore to list/get apps metadata instead of retrieving over the wire.

### `monster.appsStore`
Since the appstore list represents apps available for the account that originally logged in, this is now stored inside the `auth` module at it is based on the authentication session.

It also becomes a list as it shouldn't be accessed manually anymore (either for lookup or update). Updates are now restricted to the `auth` module, triggered by message bus events to keep local data in sync., while access is restricted to `monster.util#getAppStoreMetadata` and `monster.util#listAppStoreMetadata`.

### `auth.currentAppsStore.fetched|updated|deleted`
Listens for requests listing/modifying the current appstore and update our local copy accordingly. We do this to keep our local copy as up-to-date as possible, avoiding a round trip when app metadata is needed.

Those events potentially trigger `auth.currentUser.updated` when current user's `appList` needs to be updated, to keep current user's `appList` up-to-date.

### `auth.currentUser.updated`
Similar to previous appstore events, but for current user updates. It refreshes our local copy of current user as well as cookies locale when necessary.

----
Related PRs:
- https://github.com/2600hz/monster-ui-accounts/pull/169
- https://github.com/2600hz/monster-ui-service-plan-override/pull/48
- https://github.com/2600hz/monster-ui-service-planner/pull/40

### DIFF
--- a/src/apps/apploader/app.js
+++ b/src/apps/apploader/app.js
@@ -25,6 +25,7 @@ define(function(require) {
 		},
 
 		subscribe: {
+			'auth.currentUser.updated': '_destroyAndMaybeReopen',
 			'myaccount.closed': 'onMyaccountClosed',
 			'apploader.destroy': '_destroy',
 			'apploader.show': 'render',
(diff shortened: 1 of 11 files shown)


guillegr123 (changes requested):
Looking great, it will be nice to have all these methods standardized.

I just have a couple of small comments so far. Also, I think the ticket number in the title of the PR should be **MSTR-197** instead of **MSPB-197**, right?

azefiel:
> Looking great, it will be nice to have all these methods standardized.

Yep, it's about damn time!

> Also, I think the ticket number in the title of the PR should be **MSTR-197** instead of **MSPB-197**, right?

Absolutely, muscle memory's not always your friend 😅 !

src/apps/apploader/app.js
```diff
@@ -85,77 +86,85 @@ define(function(require) {
...
+							monster.util.listAppLinks()
+						])
+						.flatten()
+						.keyBy('id')
+						.value(),
+					appList = _
+						.chain(_.get(monster.apps, 'auth.currentUser.appList', []))
+						.map(_.partial(_.ary(_.get, 2), actionsPerId))
+						.reject(_.isUndefined)
+						.value(),
```
guillegr123, line 133: Should the app links also be listed as part of the user apps? It seems that it was not required in the previous implementation.

guillegr123:
I got another question :sweat_smile: Will the documents for `monster.util` and `monster.ui` will be updated in another PR?

src/apps/auth/app.js
```diff
@@ -1611,6 +1560,156 @@ define(function(require) {
...
+						newActionIdsList,
+						validCurrentActionIdsList
+					])
+					.flatten()
+					.reject(_.isUndefined)
+					.uniq()
+					.value();
+
+			if (
+				!_.has(monster.apps, 'auth.currentUser')
```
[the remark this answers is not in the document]
azefiel: It sure is!

src/js/lib/monster.apps.js
```diff
@@ -321,6 +326,46 @@ define(function() {
...
+							if (params.data.accountId === _.get(monster.apps.auth, monster.util.isMasquerading() ? 'originalAccount.id' : 'accountId')) {
+								successCallback = function(data, status) {
+									monster.pub('auth.currentAppsStore.fetched', {
+										response: data.data,
+										callback: function() {
+											params.success && params.success(data, status);
+										}
+									});
+								};
+							}
```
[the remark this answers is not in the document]
azefiel: I considered doing so at an earlier step during my iteration but it seemed too involved then.

Looks pretty straightforward now!

src/js/lib/monster.apps.js
```diff
@@ -404,18 +436,19 @@ define(function() {
...
-					})
+					.chain(app.name)
+					.thru(monster.util.getAppStoreMetadata)
+					.get('extensions', [])
+					.reject(_.overSome(
+						isAppAlreadyLoaded,
+						_.flow(
+							monster.util.getAppStoreMetadata,
+							_.negate(isUserPermittedApp)
+						)
```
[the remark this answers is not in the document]
azefiel: Not sure of the benefits of this, as the resulting function's name doesn't give me a clear understanding of what it does vs current composition.

src/apps/apploader/app.js
azefiel: If you look closely, they were indeed [insert ](https://github.com/2600hz/monster-ui/blob/master/src/apps/apploader/app.js#L108)at an earlier point, and we definitely want to display them to the end user.

azefiel:
> I got another question 😅 Will the documents for `monster.util` and `monster.ui` will be updated in another PR?

Not sure there is any new method to document in `monster.ui`? As for `monster.util`, we don't really want third-party devs using those tbh.

src/apps/apploader/app.js
guillegr123: Sorry, what I meant is that the app links are not displayed in the app menu in this branch because they are filtered out for not being present in `monster.apps.auth.currentUser.appList`. I have tried this setup in config.js:
```javascript
whitelabel: {
	// ...
	appLinks: {
		'http://www.google.com': {
			icon: 'https://upload.wikimedia.org/wikipedia/commons/5/53/Google_%22G%22_Logo.svg',
			i18n: {
				'en-US': {
					label: 'Google',
					description: 'Google home page'
				}
			}
		}
	}
}
```
And the app link is listed in the `master` branch, but not in `MSTR-197`.

azefiel: Good call, with the new events triggering current appstore/user refresh, fetching apps metadata needs to happen AFTER authenticated entities are retrieved, as that's when the app list is potentially updated/noramlized.

guillegr123 (approved):
Looks good to me!

PR #798, opened 2020-11-04
MCON-11: Disable participants invite for inactive conferences

azefiel:


src/apps/common/submodules/conferenceViewer/conferenceViewer.js
```diff
@@ -246,67 +246,71 @@ define(function(require) {
...
-			formattedData.conference.disabledActions = formattedData.conference.participants.length + formattedData.conference.moderators.length === 0;
-
-			return formattedData;
+				users = self.conferenceViewerFormatUsers(data.participants),
+				moderators = _.filter(users, { is_moderator: true }),
+				participants = _.filter(users, { is_moderator: false }),
+				hasUsers = _.some([
+					moderators,
+					participants
+				], _.negate(_.isEmpty));
```
guillegr123, line 255: Is it possible to only check if the participants list is not empty?
```suggestion
				hasUsers = _.some(users);
```

azefiel: Good catch, checking `users`'s emptiness would indeed be more straightforward!

azefiel:
Also updated the moderator/participants resolver to exactly match previous behavior (as not certain non-moderator necessarily have an `is_moderator` prop).

src/apps/common/submodules/conferenceViewer/conferenceViewer.js
```diff
@@ -246,67 +246,72 @@ define(function(require) {
...
+				),
+				users = self.conferenceViewerFormatUsers(data.participants),
+				hasUsers = !_.isEmpty(users),
+				moderators = _.filter(users, isModerator),
+				participants = _.reject(users, isModerator);
+
+			return {
+				backButton: args.backButton,
+				conference: _.merge({
+					disabledActions: !hasUsers,
```
guillegr123, line 261: This value does not seem to be read anywhere within this repository. Will it be used in the future?

azefiel: After do [looking into it](https://github.com/2600hz/monster-ui/commit/572e49cfc61898c5ecc81a1c95054dbfb104708f), it does seem like it being used.

guillegr123 (approved):
Looks good to me!

PR #810, opened 2020-11-11
Define monster-related members during monsterization

azefiel:


gerardozaratem (approved):
Tested the changes with an existing app and with an app created from the new skeleton.
Old behavior: Apps need to define the properties and functions to work (css, i18n, init and load)
New behavior: Apps can work without defining the specified properties and function, defaults are set correctly

PR #811, opened 2020-11-11
Add keyValueSelector UI component

azefiel:
The goal of this PR is to abstract away the concept of selecting a kay/value pair into a reusable component, such as:

keyValueSelector

This component allows the addition of new key/value while enforcing uniqueness against an predetermined map of existing pairs.

Most of the logic/styling was lifted from this [submodule](https://github.com/2600hz/monster-ui-service-plan-override/tree/master/submodules/catItemSelector).


docs/monster-ui/keyValueSelector().md
```diff
@@ -0,0 +1,97 @@
...
+    ],
+    meats: [
+        'veal', 'bison', 'chicken', 'beef', 'pork'
+    ]
+};
+
+monster.ui.keyValueSelector({
+    onSelect: function(result) {
+        window.alert('category: ' + result[0] + '\nitem: ' + result[1]);
+    },
```
guillegr123, line 71: The `onSelect` function in this example needs to be updated, to reflect that the selected category and item are received as two separate parameters:
```suggestion
    onSelect: function(category, item) {
        window.alert('category: ' + category + '\nitem: ' + item);
    },
```

src/css/lib/partials/_monster-keyValueSelector.scss
```diff
@@ -0,0 +1,70 @@
...
+		display: flex;
+		flex-direction: column;
+
+		.form-header {
+			display: flex;
+			padding: 10px 20px 0;
+
+			& > fieldset {
+				flex: 1 1 0;
+
```
guillegr123, line 17: This empty line can be removed to keep the document formatting.
```suggestion
```

src/js/lib/monster.ui.js
```diff
@@ -3468,6 +3468,223 @@ define(function(require) {
...
+			$container = monster.ui.dialog($template, _.merge({
+				autoScroll: false
+			}, _.pick(i18n, [
+				'title'
+			]))),
+			$form = $container.find('#key_value_selector_form'),
+			itemsPerCategory = _.mergeWith(
+				_
+					.chain(existing)
+					.keyBy(function(v, category) {
```
guillegr123, line 3514: `lodash#keyBy()` iteratee function only receives one parameter, so `category` is always `undefined`. I think this function should be replaced by `lodash#mapKeys()`.

guillegr123 (changes requested):
This selector looks great! There are only a couple of small changes that I added below.

src/js/lib/monster.ui.js
azefiel: Good catch! 💪

guillegr123 (approved):
Looks good to me!

PR #821, opened 2020-11-18
[MCCP-343] Fix issue where user is able to click the button twice in the last step

pcandia:


src/apps/common/submodules/navigationWizard/navigationWizard.js
```diff
@@ -211,6 +211,16 @@ define(function(require) {
 					.on('click', function(event) {
 						event.preventDefault();
 
+						$this = $(this);
+
+						if ($this.hasClass('disabled')) {
+							return;
+						}
```
azefiel, line 218: How about modifying your target selector to NOT match elements with the `disabled` class instead?
```js
.find('selector:not(.disabled)')
```

azefiel (approved):
Looks good to me!

PR #826, opened 2020-11-19
Implement cache busting on assets and modules

azefiel:
Testing this will require working with production builds as well as stubbing out a VERSION file for both the framework and the app(s) being tested.

Here's an example from sandbox:

```
  5.0-9
  refs/tags/5.0.9
  4ba264700c09207b068292b7db70599842fcc921
  2021-01-25T18:27:30Z
  https://api.github.com/repos/2600hz/monster-ui-callflows/tarball/5.0.9
```

The version number (first line) can be updated to simulate an upgrade, which should force the browser into fetching files from the server instead of relying on its cache.

Note: graphic assets are not supported atm


gerardozaratem (approved):
Tested the following scenarios with the app deployed in a virtual machine:

1. Deployed app with gulp build-prod
2. Loaded the app in the browser (CSS files show status 200 and specific transferred size)
3. Reloaded the app, CSS files now how Status 304 and no transferred size (cached)
4. Change VERSION file
5. Reloaded the app and CSS files show status 200 and a transferred size
6. Verified request of the VERSION file

No comments regarding the code

guillegr123 (changes requested):
The styles are not loaded when running on development mode. Maybe it is related to `monster.css` :thinking:

src/js/lib/monster.js
```diff
@@ -474,42 +474,37 @@ define(function(require) {
...
+					.chain(results.version)
+					.thru(monster.parseVersionFile)
+					.get('version')
+					.defaultTo(null)
+					.value();
 
-				globalCallback && globalCallback(monster.config.developerFlags.build);
+				callback && callback(monster.config.developerFlags.build);
 			});
 		},
```
guillegr123, line 509: Upon further testing, it seems that the build config is not being loaded properly. Peeking at the `results` variable, I get this object:

image

```json
{
  "version": [
    "4.3.0",
    "success",
    {
      "readyState": 4,
      "responseText": "4.3.0",
      "status": 200,
      "statusText": "OK"
    }
  ],
  "buildFile": [
    {
      "type": "production",
      "preloadApps": [
        "accounts",
        "appexchange-submission",
        "apploader",
        "appstore",
        "auth",
        "blacklists",
        "branding",
        "call-recording",
        "callerid-manipulation",
        "callflows",
        "callqueues",
        "callqueues-pro",
        "callthru-us",
        "cluster",
        "common",
        "conferences",
        "core",
        "csv-onboarding",
        "desktop-app",
        "dynamic-callerid",
        "mobile",
        "myaccount",
        "numbers",
        "port",
        "service-plan-override",
        "service-planner",
        "userportal",
        "voicemails",
        "voip",
        "webhooks"
      ]
    },
    "success",
    {
      "readyState": 4,
      "responseText": "{\"type\":\"production\",\"preloadApps\":[\"accounts\",\"appexchange-submission\",\"apploader\",\"appstore\",\"auth\",\"blacklists\",\"branding\",\"call-recording\",\"callerid-manipulation\",\"callflows\",\"callqueues\",\"callqueues-pro\",\"callthru-us\",\"cluster\",\"common\",\"conferences\",\"core\",\"csv-onboarding\",\"desktop-app\",\"dynamic-callerid\",\"mobile\",\"myaccount\",\"numbers\",\"port\",\"service-plan-override\",\"service-planner\",\"userportal\",\"voicemails\",\"voip\",\"webhooks\"]}",
      "status": 200,
      "statusText": "OK"
    }
  ]
}
```

guillegr123 (changes requested):
I was not able to test this feature, because it seems that the build configurations are not being loaded properly.

src/js/lib/monster.js
gerardozaratem: Hi @guillegr123 are you testing in a local env? I ran my tests in an already deployed for prod env and the VERSION file was updated properly. Will try to run this again.

gerardozaratem (commented):
Will retest in a different environment

src/js/lib/monster.js
```diff
@@ -267,8 +267,8 @@ define(function(require) {
 			}
 		},
 
-		css: function(href) {
-			$('<link/>', { rel: 'stylesheet', href: monster.util.cacheUrl(href) }).appendTo('head');
+		css: function(href, app) {
```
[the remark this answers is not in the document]
azefiel: Good find, just forced pushed to fix that.

src/js/lib/monster.js
guillegr123: Hi @gerardozaratem Yes, I was testing in a local env, with `gulp serve-prod`

azefiel: Interesting, I've extracted parsing steps into their own function to apply them during the success handler.

Seems to do the the trick.

guillegr123 (approved):
Looks good to me! It doesn't include app Javascript sources, only CSS/JSON/HTML files, right?

gerardozaratem (approved):
Tested the same scenarios as before with gulp serve-prod. Also tested serve-dev to check if app was loading properly.

azefiel:
@guillegr123 

Only CSS/JSON for now, as HTML files get compiled and concatenated into JS (at least for prod).

For JavaScript files, requirejs takes care of [busting](https://github.com/2600hz/monster-ui/blob/master/src/js/main.js#L112), but next step would be use the same mechanism with app version.

PR #859, opened 2020-12-23
[5.0] MPOR-9: Fix account name validation

masmerino13:


azefiel:
Actually, closing this for now, as the new port request creation wizard was reverted on 5.0 until it can be tested against Phonebook (proxied by kazoo).

PR #861, opened 2020-12-23
MACT-91: Use disabled prop for wizard done button

guillegr123:
Use the `disabled` property instead of a class to disable the navigation wizard's done button. This in turn allows the done button to be re-enabled when the step is rendered again, which solves the issue of it remaining disabled even after changing the account name.

### DIFF
--- a/src/apps/common/submodules/navigationWizard/navigationWizard.js
+++ b/src/apps/common/submodules/navigationWizard/navigationWizard.js
@@ -612,11 +612,11 @@ define(function(require) {
 				};
 
 			if (!result.valid) {
-				//If it fails for any reason then re-enable the button
+				//If validation fails for any reason then re-enable the button
 				wizardArgs
 					.container
 						.find('#done')
-						.removeClass('disabled');
+						.prop('disabled', false);
 				return;
 			}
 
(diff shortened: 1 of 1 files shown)


src/apps/common/submodules/navigationWizard/navigationWizard.js
```diff
@@ -207,19 +207,23 @@ define(function(require) {
 					});
 
 			$template
-				.on('click', '#done:not(.disabled)', function(event) {
-					event.preventDefault();
+				.find('#done')
```
azefiel, line 210: Checks against disabled pseudo-classes can be achieve directly at the selector level:
```suggestion
				.find('#done:enabled')
```

azefiel (commented):
Nice self contained commits, now if only GitHub would allow us to review them independently...

src/apps/common/submodules/navigationWizard/navigationWizard.js
guillegr123: Yes, but the element is looked for once before it is disabled. However, the click event is not triggered if the button is disabled, so I removed the extra check in the handler function.

azefiel (approved):

PR #865, opened 2021-01-07
MSTR-213: Fix app user-permission selection

azefiel:


gerardozaratem (approved):
Replicated the issue on local env with master branch
Switched to MSTR-213 branch. Issue is fixed.
Tested "Admins only" and "All users" options 
Tested Selecting all users and selecting none
Tested the process with different apps
Permission were correctly granted to the users selected

PR #872, opened 2021-01-22
MSPB-167: Fix: SmartPBX does not set account emergency caller ID when main number is selected

masmerino13:
Hi @azefiel ,

What I found here is that the emergency number isn't assigned to the account when the 911 number is added from the numbers app.

Any comment suggestion are welcome!

### DIFF
--- a/src/apps/common/i18n/en-US.json
+++ b/src/apps/common/i18n/en-US.json
@@ -636,6 +636,29 @@
 		"__version": "4.1",
 		"numbersCarrierModule": {
 			"custom": "Custom"
+		},
+		"updateCallerIdDialog": {
+			"title": {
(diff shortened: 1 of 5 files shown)


azefiel:
Not sure we want to extract auto assignment of emergency info behavior outside of SmartPBX.

I think the ask here is similar to how caller ID auto assignment is [handled](https://github.com/2600hz/monster-ui-voip/blob/master/submodules/strategy/strategy.js#L1106).

When one or several numbers are added as main numbers for an account, we should automagically setup caller ID emergency number if none are set.

If only one number has e911 setup, that number is assigned as the account's cid emergency number. If more than one number has e911 setup, we prompt the user to select which number they want to use as the account's cid emergency number.

Please let me know if that does not make sense.

PR #878, opened 2021-02-03
MBRA-12: Remove Google Plus for list of social links

azefiel:


gerardozaratem (changes requested):
Tested with the following social attribute in the config file:
image

And obtained the following result
image
image

May I suggest changing the `isSupportedSocial` function to something like this?
image

image

image

azefiel:
Nice find!

Looks like the issue comes from `lodash#filter(value, index)` feeding `index` to `lodash#get(object, path, defaultValue)` as default  value when `socialIcon` is missing.

I've modified the arity of `lodash#get` so that it only accepts two parameters.

Thanks for the review 😉

gerardozaratem (approved):
Tested the same as before. Looks good to me!

PR #880, opened 2021-02-09
MSTR-215: Fix timezone label

ramandeepromana:


pcandia (approved):
Looks good to me, i tested in accounts and it works as expected.

src/js/lib/monster.timezone.js
```diff
@@ -662,9 +662,9 @@ define(function(require) {
 
 			$.each(self.list, function(i, data) {
 				if (selected === data) {
-					dropdown.append('<option value="' + data + '" selected>' + data + '</option>');
+					dropdown.append('<option value="' + data + '" selected>' + data.replace(/_/g, ' ') + '</option>');
 				} else {
-					dropdown.append('<option value="' + data + '">' + data + '</option>');
+					dropdown.append('<option value="' + data + '">' + data.replace(/_/g, ' ') + '</option>');
 				}
 			});
```
azefiel, line 669: When an operation is repeated more than once, it's a clue it should be abstracted into it's own routine/variable etc.

In our case, the result of `string#replace()` could be stored into a variable that could then be used to refer to the human readable version of the string:

```js
var humanReadableLabel = data.replace(/_/g, ' ');
```

One more improvement can be made in that block to reduce duplication. It is not related to the purpose of this PR, but it's small enough that we can pack it up with it as well:

```js
dropdown.append(
    '<option value="' + data + '"' + selected === data ? 'selected' : '' + '>' + humanReadableLabel + '</option>'
);
```

Now this is a bit long so we can construct the `option`  tag programmatically as a jQuery object:

```js
var humanReadableLabel = data.replace(/_/g, ' '),
    $option = $('<option>', {
        value: data,
        selected: selected === data,
        text: humanReadableLabel
    });

dropdown.append($option);
```

Okay, that's better! If we happen to zoom out a bit to the outer block, we can see the operation of creating an `option` tag is repeated later on. Let's extract it into it's own function so we go away with the repeated logic!

```js
var self = this,
	selected = _selected || self.getLocaleTimezone(),
	getOptionTag = _.partial(function(selected, args) {
		var value = args.value;

		return $('<option>', {
			value: value,
			selected: selected === value,
			text: args.text
		});
	}, selected);
```

Once that's done, we use our newly created function instead of manually creating `option` tags:

```js
var self = this,
	selected = _selected || self.getLocaleTimezone(),
	getOptionTag = _.partial(function(selected, args) {
		var value = args.value;

		return $('<option>', {
			value: value,
			selected: selected === value,
			text: args.text
		});
	}, selected);

$.each(self.list, function(i, data) {
	var humanReadableLabel = data.replace(/_/g, ' ');

	dropdown.append(getOptionTag({
		value: data,
		text: humanReadableLabel
	}));
});

if (extraOptions) {
	_.each(extraOptions, function(name, value) {
		dropdown.prepend(getOptionTag({
			value: value,
			text: name
		}));
	});
}
```

And voilà, we've both added a fix and improved the function's structure responsible for generating the dropdown!

ramandeepromana: @azefiel Thanks for the awesome explanation and I've noted everything for the future ✌🏼

azefiel: Feel free to implement those suggestions using what you've learn from how to make atomic commits!

pcandia (approved):
Looks great, I tested it and verified that it works as expected.

PR #887, opened 2021-02-12
MSTR-218: Expose service plan item `name` prop as editable

azefiel:
service-plan-item-name


pcandia (approved):
looks good to me, i did testing and it was working as expected.

PR #888, opened 2021-02-12
[5.0] MSTR-218: Expose service plan item `name` prop as editable

azefiel:


pcandia (approved):
looks good to me, i did testing and it was working as expected.

PR #889, opened 2021-02-12
MSTR-214: Validate list of space separated e911 email addresses

azefiel:
listOf-validator

### DIFF
--- a/src/apps/core/i18n/en-US.json
+++ b/src/apps/core/i18n/en-US.json
@@ -144,7 +144,10 @@
 		"defaultRules": {
 			"required": "This field is required.",
 			"remote": "Please fix this field.",
-			"email": "Please enter a valid email address.",
+			"email": {
+				"one": "Please enter a valid email address.",
+				"list": "Please enter valid space separated email addresses."
+			},
 			"url": "Please enter a valid URL.",
 			"date": "Please enter a valid date.",
 			"dateISO": "Please enter a valid date (ISO).",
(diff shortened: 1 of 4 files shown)


src/apps/core/i18n/en-US.json
```diff
@@ -144,7 +144,10 @@
 		"defaultRules": {
 			"required": "This field is required.",
 			"remote": "Please fix this field.",
-			"email": "Please enter a valid email address.",
+			"email": {
+				"one": "Please enter a valid email address.",
```
guillegr123, line 148: The e-mail rule seems to be broken when a single e-mail is validated:
image

guillegr123 (changes requested):

src/js/lib/monster.ui.js
```diff
@@ -1280,7 +1280,35 @@ define(function(require) {
...
+						.thru(getRuleI18n)
+						.get(plural)
+						.value();
+				},
+				getRuleMessageForOne = _.flow(
+					_.over([
+						_.partial(getRuleMessageForPlural, 'one'),
+						getRuleI18n
+					]),
+					_.partial(_.find, _.isString)
```
guillegr123, line 1310: In this command, the first parameter for `lodash#find` needs to be the array to search over:
```suggestion
					_.partial(_.find, _, _.isString)
```

guillegr123 (changes requested):

src/js/lib/monster.ui.js
```diff
@@ -1280,7 +1280,35 @@ define(function(require) {
...
+						.chain([
+							'customRules',
+							'defaultRules'
+						])
+						.map(_.partial(_.ary(_.get, 2), i18n))
+						.find(_.flow(
+							_.partial(_.ary(_.get, 2), _, 'ruleId'),
+							_.negate(_.isUndefined)
+						))
+						.value();
```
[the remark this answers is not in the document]
azefiel: The stream was refactored so that path resolution happens before the truthy check (to avoid accessing the same path twice).

guillegr123 (approved):
Looks good to me!

PR #892, opened 2021-02-15
OOB: Fix context binding for regex rule methods

guillegr123:


masmerino13 (approved):
I don't have comments to add

PR #894, opened 2021-02-17
MSPB-172: Add support for 'Sugar' library to format dates for Office …

pcandia:
…Holidays


azefiel (commented):
Since we're only interested in the `date` parsing functionality right now, have you looked at [customizing](https://sugarjs.com/download/#/Custom) the build to contain only that module?

pcandia:
Yes, good point! Let me do that

azefiel (commented):
Mind using `sugar-date` as module id, to match the npm module name?

It'd make sense for the file to match that name as well, since it only contains that specific module.

azefiel (commented):
`sugar.min.js` is now included along `sugar-custom.min.js`, I believe we're only interested in the custom build, right?

pcandia:
yes you are right! 👍

azefiel (approved):

PR #897, opened 2021-02-19
[5.0] Fix pro app build

azefiel:


gerardozaratem (commented):
@azefiel Some of these changes were already on master branch since the implementation of cache busting, right?

azefiel:
Exactly, that's why the diff count is higher than for the master PR.

PR #904, opened 2021-03-18
MPRO-29: Extending monster ui mask for ipv4 and to accept other custom patterns

gerardozaratem:
ipv4 pattern is now a part of the presets
Accepts other patterns if the specified type is not a preset, uses options the same as jquery mask

### DIFF
--- a/src/js/lib/monster.ui.js
+++ b/src/js/lib/monster.ui.js
@@ -2685,8 +2685,8 @@ define(function(require) {
 			self.onNavbarTabClick(thisArg, $tab, oArgs);
 		},
 
-		mask: function(target, type) {
-			var validations = {
+		mask: function(target, type, options) {
+			var config = _.get({
 				phoneNumber: {
 					mask: 'AZZZZZZZZZZZZZZZZ',
 					options: {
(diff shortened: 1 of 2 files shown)


src/js/lib/monster.ui.js
```diff
@@ -2712,6 +2712,16 @@ define(function(require) {
 						}
 					}
 				},
+				ipv4: {
+					mask: '0ZZ.0ZZ.0ZZ.0ZZ',
+					options: {
+						translation: {
+							'Z': {
+								pattern: /[0-9]/, optional: true
```
azefiel, line 2720: Please set each prop on a new line for readability's sake.

azefiel (commented):
Looks good, let's take the opportunity to refactor the method with declarative style and get rid of `if/else` statement/manual existence check:
```js
function mask($target, type, options) {
  var config = _.get({
    phoneNumber: {},
    macAddress: {},
    ipv4: {},
    extension: {},
  }, type, {
    mask: type,
    options: options
  });

  $target.mask(config.mask, config.options);
}
```
Also, don't forget to update the related [doc entry](https://github.com/2600hz/monster-ui/blob/master/docs/monster-ui/mask().md), here's an [example](https://docs.2600hz.com/ui/docs/monster-util/toFriendlyDate%28%29/) on how to display shortcuts and optional parameters (let me know if you have any question on this).

azefiel (approved):
Looks good, merge away!

PR #905, opened 2021-03-18
MSPB-183: Add support for 'date-holidays' library to be able to impor…

pcandia:
…t national holidays worldwide


azefiel (commented):
We need to support at least IE11, which, according to the [library's browser support table](https://github.com/commenthol/date-holidays#browser) needs es6 polyfills.

A [shim](https://requirejs.org/docs/api.html#config-shim) entry will have to be defined so the polyfills are loaded before the library.

pcandia:
I will be removing the last commit as we don't need to worry about the support for IE11

azefiel (approved):
Looks good, merge away!

PR #913, opened 2021-04-27
MOCP-4: Check if app is pro, if it is then append '-pro' to the app name

pcandia:


azefiel (commented):
That's a solid start!

One major problem though, is the metadata file being completely overridden by an object only containing the `name` prop. That's problematic, as all the config information required for the app to be loaded in the DB correctly is missing 😱

What needs to happen is read the contents of the existing metadata file, update the `name` prop, and then write that back to the file.

Similarly to how we use `fs#writeFileSync`, node provides [`fs#readFileSync`](https://nodejs.org/dist/latest-v15.x/docs/api/fs.html#fs_fs_readfilesync_path_options) to read files. The output would then need to be parsed to obtain a regular JS object we can operate on (in this case, modify the `name` prop).

Since we are using the synchronous methods to manipulate files, we need to make sure to only do it when needed.

Also, the name value from the metadata file should be used directly, rather than the one passed in the CLI.

Let me know if you need more clarifications, we could potentially pair on this!

PR #924, opened 2021-07-27
[5.0] SPRO-103: Make footable page sizes customizable

gerardozaratem:
* Template used to re add paging element when footable function pageSize removes it

* Changing template to render the specified page sizes instead of hardcoded sizes

* Adding default available page sizes

* Expaning functionality to accept customs available page sizes and fixing the issue of paging element disapperaing

* Adding next and prev to paging template

* Sending row count to paging template

* Removing unused pagesizes labels from translation files


gerardozaratem:
closing PR

PR #928, opened 2021-07-27
MOCP-4: Override metadata with pro version for pro builds

azefiel:
The goal of this PR is for the build system to output a different `metadata.json` for apps built with the pro vs standard switch (`--pro`).

Reference [doc entry](https://github.com/2600hz/monster-ui/blob/master/docs/building.md) for build commands.


azefiel:
@gerardozaratem ^^

pcandia (approved):
no comments to add, looks good to me!

PR #936, opened 2021-08-31
Update node packager

azefiel:
- Update node packager
- Rebuild node sass bindings on new environment


masmerino13 (approved):
I don't have any comments to add

PR #975, opened 2021-11-17
SPRO-131: Adding support for Alcatel models

gerardozaratem:
New models and new brand added.
Brand added: Alcatel
Models added: Myraid M3, Myraid M5, Myraid M7, Halo H3G, Halo H6


azefiel (approved):
Looks good, merge away!

PR #976, opened 2021-11-24
[5.0] SPRO-78: Adds Grandstream HT8xx

plau:


vlad88sv:
@plau could you please re-send it with a fixed name in the branch name? it has  `5.0]SPRO-78 `right now and that may affect the build process. Thank you

PR #983, opened 2021-12-07
SPRO-131: Adding image for Alcatel Halo H3P model

gerardozaratem:
Adding image for additional Alcatel model (Halo H3P)


pcandia (approved):
lgtm, no comments to add

PR #991, opened 2022-01-24
SPRO-107: adds images for missing grp26xx models

plau:


azefiel (commented):
Just wanted to make sure that 5.0 was open for non-fix merges?

PR #997, opened 2022-03-07
SPRO-131: Updating images for models in the ALE Myriad family 

gerardozaratem:
Updating images for model Myriad M3, M5 and M7


azefiel (approved):
Nice catch, realizing I didn't notice the typo on the original review either 🙏

PR #1000, opened 2022-04-08
AXS-9: add a new section to app exchange to connect cluster to marketplace

icehess:
This adds a new section to app exchange to link the cluster to marketplace and configure the kazoo_appex.


icehess:
Required PRs:
* https://github.com/2600hz/kazoo-crossbar/pull/261
* https://github.com/2600hz/kazoo-core/pull/382

src/apps/appstore/app.js
```diff
@@ -393,8 +471,296 @@ define(function(require) {
...
+							name: formData.cluster_name
+						},
+						function() {
+							self.showMarketplaceConnector(
+								parent
+							);
+						},
+						function() {
+							saveThis.prop({ 'disabled': false });
+						});
```
azefiel, line 550: One thing to note here, this code is executed asynchronously, meaning `self#updateMarketConnector` might start executing BEFORE `maybeSetApiUrl` is done executing.

If `self#updateMarketConnector` relies on `maybeSetApiUrl` being executed before it starts its own API request, these requests will have to be handled with a control flow.

Here, we basically want to run them in series (waiting for one to resolve before the next one starts):
```js
monster.series([
   maybeSetApiUrl,
   function(next) {
      self.updateMarketConnector(..., _.partial(next, null), _.partial(next, true))
   }
], function(err) {
   if (err) {
      saveThis.prop({ 'disabled': false });
      return;
   }
   self.showMarketplaceConnector(parent);
});
```
And `maybeSetApiUrl` will have to be modified to take in a callback:
```js
			const maybeSetApiUrl = function(next) {
				if (!self.appstoreData.marketplace.api_url && monster.config.api.default) {
					self.updateMarketConnector({
						action: 'api_url',
						api_url: monster.config.api.default
					}, _.partial(next, null));
				}
			};
```

src/apps/appstore/app.js
```diff
@@ -393,8 +471,296 @@ define(function(require) {
...
+		bindMarketWelcomeEvents: function(parent) {
+			const self = this;
+			const maybeSetApiUrl = function() {
+				if (!self.appstoreData.marketplace.api_url && monster.config.api.default) {
+					self.updateMarketConnector({
+						action: 'api_url',
+						api_url: monster.config.api.default
+					});
+				}
+			};
```
azefiel, line 710: Looks like this logic is used in more than one place, os it might be a good idea to extract it in its own method.

azefiel:
@icehess Let me know if you wanna go through those "live" like you proposed.

icehess:
@azefiel Thanks for your wonderful comments. I learn a good amount about Monster UI now!

I hope I addressed everything you mentioned.

src/apps/appstore/app.js
```diff
@@ -13,19 +13,36 @@ define(function(require) {
 			'fr-FR': { customCss: false },
 			'ru-RU': { customCss: false }
 		},
+		requests: {
+			'marketplace.get': {
+				apiRoot: monster.config.api.default,
+				// generateError: false,
+				url: 'apps_store/marketplace',
```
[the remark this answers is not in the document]
azefiel: My bad, by private I meant: does it make sense for third-party developers to access this endpoint?

Hence exposing it in the SDK or keeping it "hidden" inside the app.

src/apps/appstore/app.js
```diff
@@ -13,18 +13,74 @@ define(function(require) {
...
+				hasValue = _.toArray(arguments).length === 2,
+				store = ['_store'];
+			_.set(
+				self,
+				hasValue
+					? _.flatten([store, _.isString(path) ? path.split('.') : path])
+					: store,
+				hasValue ? value : path
+			);
+		},
```
azefiel, line 62: Just making it clear that those methods are [exposed by default](https://github.com/2600hz/monster-ui/blob/5.1/src/js/lib/monster.apps.js#L436) on both `master` and `5.1`.

Copy/pasting them outright should only be done if there is plan to cherry-pick this PR's changes into 5.0.

Even then, there is point to be made so that changes targeting 5.1 use 5.1+ features and if backporting is needed, that port gets adapted to whatever version it targets, but that's just my opinion.

src/apps/appstore/app.js
```diff
@@ -13,18 +13,74 @@ define(function(require) {
...
 			var self = this,
-				template = $(self.getTemplate({
-					name: 'app'
-				})),
 				parent = container || $('#monster_content');
 
+			self.setStore({
+				shouldShowMarket: monster.util.isSuperDuper() && !monster.util.isMasquerading(),
+				appstoreData: null
+			});
```
azefiel, line 71: 🙌

src/apps/appstore/app.js
```diff
@@ -38,9 +55,12 @@ define(function(require) {
...
-		bindEvents: function(parent, appstoreData) {
+		appstoreData: null,
+
+		bindEvents: function(parent, appstoreData, shouldShowMarket) {
 			var self = this,
-				searchInput = parent.find('.search-bar input.search-query');
+				searchInput = parent.find('.search-bar input.search-query'),
+				appListContent = null;
 
 			setTimeout(function() { searchInput.focus(); });
```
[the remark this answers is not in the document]
azefiel: I don't have to click anywhere actually, basically:
Apploader > "App Exchange" > Apploader > "App Exchange" > error is logged in the console

Also, this only happens when not using your mock data, and the first request throws an error.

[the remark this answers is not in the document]
azefiel: Highly likely that the detach is responsible. The way that focus is attached though (with `setTimeout`) is dirty anyways, should be done once the template is in the DOM. Will try that out tomorrow.

src/apps/appstore/app.js
```diff
@@ -51,15 +109,27 @@ define(function(require) {
 				parent.find('.app-filter').removeClass('active');
 				$this.addClass('active');
 
+				if (appListContent) {
+					parent.find('.left-menu .marketplace .launch-market').removeClass('active');
+					// Can't use monster.ui.insertTemplate because of the way
+					// we detach and fadeOut using `> *` selector.
```
azefiel, line 115: Yep yep

src/apps/appstore/app.js
```diff
@@ -393,8 +513,274 @@ define(function(require) {
...
+				e.preventDefault();
+
+				const saveThis = $(this);
+
+				saveThis.prop({ disabled: true });
+
+				if (monster.ui.valid(linkForm)) {
+					const formData = monster.ui.getFormData('cluster_link_form');
+
+					self.maybeSetApiUrl();
```
azefiel, line 565: If this potentially needs to run on every `app#updateMarketConnector`, how about moving it in that method body then?

src/apps/appstore/app.js
```diff
@@ -393,8 +513,274 @@ define(function(require) {
...
+			self.bindMarketWelcomeEvents(parent, template);
+		},
+
+		bindMarketWelcomeEvents: function(parent) {
+			const self = this;
+
+			parent.find('#market_enable_connector_btn').on('click', function(e) {
+				e.preventDefault();
+
+				this.maybeSetApiUrl();
```
azefiel, line 723: `this` in this context will refer to the current function it's in (`function(e) {}`), which does not expose any method named `maybeSetApiUrl`, and throw a TypeError.

That's why the `self` hack is used, to keep a reference to the outer function's `this`.

src/apps/appstore/app.js
```diff
@@ -102,12 +189,44 @@ define(function(require) {
...
+					if (!shouldShowMarket) {
+						return next();
+					}
+					monster.request({
+						resource: 'marketplace.get',
+						success: function(response) {
+							next(null, response.data);
+						},
+						error: function() {
+							next();
```
azefiel, line 215: Still unsure what will happen here.

Is the `/marketplace` endpoint expected to always be available, or will that be conditioned by a system flag (like `/external_numbers` for example)?

If the former, then this is fine, otherwise, maybe we can hide the `Marketplace Connector` button if this errors out?

src/apps/appstore/app.js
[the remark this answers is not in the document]
icehess: I don't see any benefits of adding to SDK, the API is just and if some super duper admin there want to do something they can use it directly. But as for 3-party developer, i don't think any MUI app wants to use this.

src/apps/appstore/app.js
```diff
@@ -102,12 +189,44 @@ define(function(require) {
...
+						}
+					});
+				}
+			}, function(err, data) {
+				if (!err) {
+					self.setStore('appstoreData', data);
 				}
-			}, callback);
+				callback(err);
```
azefiel, line 223: This is backward, if it errors out, we want to exit early (which is accomplished by the `return` statement):
```js
if (err) {
   return callback(err)
}
self.setStore(' appstoreData', data);
```

azefiel (commented):
Almost there!

src/apps/appstore/app.js
icehess: That makes a loop, `maybeSetApiUrl` also calls `updateMarketConnector` to set `api_url`.

Also the API is implement in a way that you can only perform a single action one at a time. You **_cannot_** set `enabled` and `api_url` in the same request.

If you really need those fancy update/post thing then you could always use `/system_config/appex_client` API directly. The `/appstore/marketplace` is a convenient way to configure appex_client without needing to activate `/system_config` crossbar API.

Also there is only few actions here, `link`/`unlink`, `enable` and `api_url`. We just want to make sure we set api_url when you're enabling and/or linking your cluster to make sure the appex client is functional without requiring you to set it manually when you enable and linking your cluster.

src/apps/appstore/app.js
icehess: fixed

icehess:
@azefiel I think the last two commits actually fixed the problem search input.

What I did was first revert storing appstoreData, and passing it around as before.

Then I wrapped the while app list section to a container and moved all logic about rendering and binding of search input and app list elements to its own function.

So now when we switch between marketplace connector and app lists, instead of using `detach` we just use vanilla  `monster.ui.insertTemplate` to switch between them.

I hope this   this fix the throw error you mentioned (I still cannot reproduce it).

This also fixes another issue that I had, when switch back from market connector, the search input won't filters the app. But now it is fixed!

src/apps/appstore/app.js
azefiel: Yep, that's why we usually wrap request handlers in their own methods and create yet other methods to wrap higher levels of abstraction (e.g. enableMarketConnect, linkMarktConnector, ...).

src/apps/appstore/app.js
azefiel: Yep, that suggestion was incorrect, and since we don't expect the UI to function when one of those requests errors out, there is no need to pass down that error down the callback chain.

azefiel (approved):

PR #1005, opened 2022-05-13
MSTR-228: Allow app custom routing

guillegr123:
This change allows to redirect to a registered app, if the first 2 parts of the hash route match a valid app, on load/re-load. For example:

- `apps/callcenter/queues/abc123/members` redirects to `apps/callcenter`
- `apps/cluster/tasks` redirects to `apps/cluster`

This enables Monster UI to load, reload or redirect to apps that use their own sub-routing system.

Normal routes (app/{appName}) and invalid routes should work as usual.

### DIFF
--- a/src/js/lib/monster.routing.js
+++ b/src/js/lib/monster.routing.js
@@ -104,14 +104,15 @@ define(function(require) {
 		},
 
 		addDefaultRoutes: function() {
-			this.add(/^apps\/([^/?]+)(?:\/[^?]*)?\??(.*)?$/, function(appName, queryString) {
+			this.add(/^apps\/([^/?]+)(\/[^?]*)?\??(.*)?$/, function(appName, restSegment, queryString) {
 				// not logged in, do nothing to preserve potentially valid route to load after successful login
 				if (!monster.util.isLoggedIn()) {
 					return;
 				}
 
-				// Do not reload app if it is already loaded (just the rest segment or the query has changed)
-				if (monster.apps.getActiveApp() === appName) {
+				// Do not reload app if it is already loaded and there is a rest segment,
+				// to allow the app to handle its own routing
+				if (monster.apps.getActiveApp() === appName && !_.isEmpty(restSegment)) {
 					return;
 				}
 
--- a/src/js/lib/monster.routing.js
+++ b/src/js/lib/monster.routing.js
@@ -104,11 +104,17 @@ define(function(require) {
 		},
 
 		addDefaultRoutes: function() {
-			this.add('apps/{appName}:?query:', function(appName, query) {
+			this.add(/^apps\/([^/?]+)(?:\/[^?]*)?(\??.*)?$/, function(appName, query) {
 				// not logged in, do nothing to preserve potentially valid route to load after successful login
 				if (!monster.util.isLoggedIn()) {
 					return;
 				}
+
+				// Do not reload app if it is already loaded (just the rest segment or the query has changed)
+				if (monster.apps.getActiveApp() === appName) {
+					return;
+				}
+
 				var availableApps = monster.util.listAppStoreMetadata('user');
 
 				// try loading the requested app
--- a/src/js/lib/monster.routing.js
+++ b/src/js/lib/monster.routing.js
@@ -104,7 +104,7 @@ define(function(require) {
 		},
 
 		addDefaultRoutes: function() {
-			this.add(/^apps\/([^/?]+)(\/[^?]*)?\??(.*)?$/, function(appName, restSegment, queryString) {
+			this.add(/^apps\/([a-z]+(?:-[a-z]+)?)(\/[^?]*)?\??(.*)?$/, function(appName, restSegment, queryString) {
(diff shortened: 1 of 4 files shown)


azefiel:
Unless some changes are missing in this PR, monster.routing#hasMatch does not take any arguments: it checks the current hash against apps/{appName}:?query.

guillegr123:
Sorry @azefiel , you are absolutely right, I forgot to commit the change for `monster.routing#hasMatch` 😅 

It should be up now. Thanks for the heads up.

guillegr123:
Hi @azefiel, I tried using the rest fragments, but it seems that Crossroads.js does not support having both an optional rest fragment and an optional query capturing group. So I ended up using a regular expression instead, and exposing the `parseQueryString()` function at `monster.util`, to convert the extracted query string into an object as Crossroads.js used to do.

src/js/lib/monster.util.js
```diff
@@ -1632,6 +1566,75 @@ define(function(require) {
 			.value();
 	}
 
+	/**
```
azefiel, line 1569: This is a clean cp/mv, yeah?

src/js/lib/monster.routing.js
```diff
@@ -104,11 +104,17 @@ define(function(require) {
...
 
 		addDefaultRoutes: function() {
-			this.add('apps/{appName}:?query:', function(appName, query) {
+			this.add('apps/{appName}/:rest*::?query:', function(appName, _rest, query) {
 				// not logged in, do nothing to preserve potentially valid route to load after successful login
 				if (!monster.util.isLoggedIn()) {
 					return;
 				}
+
+				// Do not reload app if it is already loaded (just the rest segment or the query has changed)
```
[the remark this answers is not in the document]
guillegr123: Yes, I tried it like `http://localhost:3000/#/apps/voip?m=<account_id>`, and it loads the app masquerading to the specified account.

src/js/lib/monster.util.js
guillegr123: Yes, it was a clean cut and paste.

src/js/lib/monster.routing.js
```diff
@@ -104,7 +104,7 @@ define(function(require) {
 		},
 
 		addDefaultRoutes: function() {
-			this.add('apps/{appName}/:rest*::?query:', function(appName, _rest, query) {
+			this.add(/^apps\/([^/?]+)(?:\/[^?]*)?(\??.*)?$/, function(appName, query) {
```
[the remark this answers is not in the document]
guillegr123: Sure!

src/js/lib/monster.routing.js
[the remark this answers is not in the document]
guillegr123: Reloading the app by clicking the app icon in the topbar was not working at all 😅 I pushed another commit to fix it. Masquerading works as well, both by passing the account ID to masquerade via query string (`http://localhost:3000/#/apps/voip?m=<account_id>`) and after clicking the topbar app icon while already masquerading.

src/js/lib/monster.routing.js
```diff
@@ -104,12 +104,20 @@ define(function(require) {
 		},
 
 		addDefaultRoutes: function() {
-			this.add('apps/{appName}:?query:', function(appName, query) {
+			this.add(/^apps\/([^/?]+)(\/[^?]*)?\??(.*)?$/, function(appName, restSegment, queryString) {
```
azefiel, line 107: Since app name characters are in the range [a-z-], what do you thing about restricting the first capturing group to that specific pattern?

`^apps\/([a-z]+(?:-[a-z]+)?)(\/[^?]*)?\??(.*)?$`

image

src/js/lib/monster.routing.js
```diff
@@ -104,12 +104,20 @@ define(function(require) {
...
 				// not logged in, do nothing to preserve potentially valid route to load after successful login
 				if (!monster.util.isLoggedIn()) {
 					return;
 				}
-				var availableApps = monster.util.listAppStoreMetadata('user');
+
+				// Do not reload app if it is already loaded and there is a rest segment,
+				// to allow the app to handle its own routing
```
azefiel, line 114: How about creating meaningful variable names instead of adding comments here, as comments don't tend to be updated when code changes?

I feel like the following conveys the same information:
```js
var isActiveApp = monster.apps.getActiveApp() === appName,
    hasCustomRouting = !_.isEmpty(restSegment);

```

azefiel (commented):
Last nitpicking, just want to make sure this code is as understandable as possible for our future self 😅

src/js/lib/monster.routing.js
guillegr123: Yes, sounds good to me, that will ensure that the app name has the right pattern 👍

src/js/lib/monster.routing.js
guillegr123: You are right, is easier to read and understand that way 😮

azefiel (approved):

PR #1011, opened 2022-08-11
MGDC-3: Pattern used in crossroads is no longer trimming app names with more than one hyphen

shark764:
…more than one hyphen

Current pattern provided to crossroads.addRouter to extract appName from url is trimming values such as `service-plan-assignment` and `integration-google-drive`, due to it stops after first hyphen, so for example `integration-google-drive` becomes `integration-google`.
Apps affected by this issue
- Service Plan Assignment
- Service Plan Override
- One Bill Integration
- Google Drive Connector

### DIFF
--- a/src/js/lib/monster.routing.js
+++ b/src/js/lib/monster.routing.js
@@ -104,7 +104,7 @@ define(function(require) {
 		},
 
 		addDefaultRoutes: function() {
-			this.add(/^apps\/([a-z]+(?:-[a-z]+)?)(\/[^?]*)?\??(.*)?$/, function(appName, restSegment, queryString) {
+			this.add(/^apps\/([a-z]+(?:-[a-z]+)*)(\/[^?]*)?\??(.*)?$/, function(appName, restSegment, queryString) {
 				// not logged in, do nothing to preserve potentially valid route to load after successful login
 				if (!monster.util.isLoggedIn()) {
 					return;


Mariotutal (dismissed):

masmerino13 (approved):
lgtm

src/js/lib/monster.routing.js
```diff
@@ -104,7 +104,7 @@ define(function(require) {
 		},
 
 		addDefaultRoutes: function() {
-			this.add(/^apps\/([a-z]+(?:-[a-z]+)?)(\/[^?]*)?\??(.*)?$/, function(appName, restSegment, queryString) {
+			this.add(/^apps\/([a-z]+(?:-[a-z]+(-[a-z]+){0,1})?)(\/[^?]*)?\??(.*)?$/, function(appName, restSegment, queryString) {
```
azefiel, line 107: This pattern won't work for app IDs with more than three hyphened words.

How about making it generic, by making the non-catching group greedy (match as need, 0 or more times):

```suggestion
			this.add(/^apps\/([a-z]+(?:-[a-z]+)*)(\/[^?]*)?\??(.*)?$/, function(appName, restSegment, queryString) {
```
download (1)

azefiel (changes requested):

src/js/lib/monster.routing.js
shark764: Thanks, I was looking for a way to make it match 0 or more but forgot about the `*`
image

Done

azefiel (approved):

Irvandoval (approved):

PR #1013, opened 2022-08-16
MNUM-12: Display info dialog when searching for a number that is used…

pcandia:
… in another account


masmerino13 (approved):
no comments to add

PR #1021, opened 2022-09-14
MSPB-295,MSPB-296: Pass the trigger element to the clipboard callback

oscbco:
Pass the element that triggers `clipboard` click events to the invoking callback in order to access their custom `dataset` attributes for copying purposes
Show a custom success message if needed but fallback to the previous one if none is provided

### DIFF
--- a/src/js/lib/monster.ui.js
+++ b/src/js/lib/monster.ui.js
@@ -3328,22 +3328,22 @@ define(function(require) {
 			}
 		},
 
-		clipboard: function(pTarget, value) {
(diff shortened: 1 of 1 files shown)


guillegr123 (approved):
Lgtm!

guillegr123 (approved):

guillegr123:
I think it would be a good chance to add the JSDoc to the `clipboard` function, as well as an entry in the [markdown document collection](https://github.com/2600hz/monster-ui/tree/master/docs/monster-ui). What do youthink @oscbco @azefiel ?

PR #1035, opened 2022-11-01
[5.1] MSPB-295,MSPB-296: Pass the trigger element to the clipboard callback

oscbco:
Pass the element that triggers clipboard click events to the invoking callback in order to access their custom dataset attributes for copying purposes
Show a custom success message if needed but fallback to the previous one if none is provided


guillegr123 (approved):

masmerino13 (approved):
lgtm

PR #1039, opened 2022-11-15
[5.0] SPRO-165: Provisioning Polycom Edge's families (100, 200, 300, 400 & 500)

jrmende5:
added images for polycom's model 100 220 300 320 350 400 450 500 550


gerardozaratem (approved):
Tested in my local environment, the images were displayed as expected.

PR #1042, opened 2022-12-15
MSPB-315: Cross-site message based routing

ramandeepromana:
Logic for cross-site message event binding, publication and authorization check as well as removal of event routing flag (_used in previous implementation_)


ramandeepromana:
Adding @guillegr123 to PR https://github.com/2600hz/monster-ui/pull/1042 and https://github.com/2600hz/monster-ui-voip/pull/475 as @joristirado is not longer available 🥲

src/apps/core/app.js
```diff
@@ -89,13 +90,35 @@ define(function(require) {
...
 				_.bind(self.loadIncludes, self)
 			], _.bind(self.loadAuth, self));
 		},
 
+		bindCrossSiteMessagingHandler: function(crossSiteMessaging) {
+			const { origin, topics } = crossSiteMessaging;
+
+			window.addEventListener('message', handler);
+
+			function handler(event) {
```
guillegr123, line 109: Declare the function before it is added as an event listener, using a more explicit name, to follow Monster UI conventions. For example:
```suggestion
			var handleCrossSiteMessages = function handleCrossSiteMessages(event) {
```

src/apps/core/app.js
```diff
@@ -76,7 +76,8 @@ define(function(require) {
 				mainTemplate = $(self.getTemplate({
 					name: 'app',
 					data: dataTemplate
-				}));
+				})),
+				crossSiteMessaging = monster.config.crossSiteMessaging;
```
guillegr123, line 80: Please add the corresponding documentation for `crossSiteMessaging` in the `configuration.md` file.

src/apps/core/app.js
```diff
@@ -89,13 +90,35 @@ define(function(require) {
...
 			], _.bind(self.loadAuth, self));
 		},
 
+		bindCrossSiteMessagingHandler: function(crossSiteMessaging) {
+			const { origin, topics } = crossSiteMessaging;
+
+			window.addEventListener('message', handler);
+
+			function handler(event) {
+				if (event.origin !== origin || !topics.includes(event.data)) {
```
guillegr123, line 110: Only a single origin will be allowed? Is it possible or necessary to allow more than one origin?

Also, another thought, what if we leave to each application to check which topics are allowed for cross site messaging?

src/apps/core/app.js
```diff
@@ -89,13 +90,35 @@ define(function(require) {
...
+					return;
+				}
+
+				if (!monster.util.isAuthorizedTopicForCrossSiteMessaging(event.data)) {
```
guillegr123, line 114: The function `isAuthorizedTopicForCrossSiteMessaging` currently does not receive any parameters.

src/apps/core/app.js
```diff
@@ -89,13 +90,35 @@ define(function(require) {
...
+					return;
+				}
+
+				monster.pub('core.crossSiteMessage', event.data);
```
guillegr123, line 118: Could there be a way to route the messages for each app, also taking into account if it is active, or not? Or maybe even adding the name of the app to the data, so for example we could call:

```suggestion
				monster.pub(event.data.app + '.crossSiteMessage', event.topic);
```

Should we also consider allowing to send some payload to the application, besides the topic?

src/js/lib/monster.util.js
```diff
@@ -81,9 +81,26 @@ define(function(require) {
...
 		unformatPhoneNumber: unformatPhoneNumber,
 		unixToDate: unixToDate,
-		updateImagePath: updateImagePath
+		updateImagePath: updateImagePath,
+		isAuthorizedTopicForCrossSiteMessaging: isAuthorizedTopicForCrossSiteMessaging
 	};
 
+	function isAuthorizedTopicForCrossSiteMessaging() {
+		var TopicForCrossSiteMessaging = [
+			'voip.tab.myOffice',
```
guillegr123, line 90: Is there a reason to use a topic name that is different to the ones that the app already handles?

guillegr123 (changes requested):

src/apps/core/app.js
ramandeepromana: Please let me know if the documentation is sufficient or if we should also include an example in there.

src/apps/core/app.js
ramandeepromana: My bad totally missed this one.

src/apps/core/app.js
ramandeepromana: Apologies for forgetting to include a check for the active/load status of the application. It has been fixed now. After reviewing the .pub and .sub documentation, I decided to use 'core.crossSiteMessage' along with the application name.

We could, but I'm really not sure, as this was just an enhancement based on @joristirado's RFC to the previous implementation. Moreover, Verizon themselves will not be using this anymore.

src/js/lib/monster.util.js
ramandeepromana: Apologies again but that was done based on RFC that @joristirado passed on.

src/js/lib/monster.util.js
```diff
@@ -81,9 +81,26 @@ define(function(require) {
 		},
 		unformatPhoneNumber: unformatPhoneNumber,
 		unixToDate: unixToDate,
-		updateImagePath: updateImagePath
+		updateImagePath: updateImagePath,
+		isAuthorizedTopicForCrossSiteMessaging: isAuthorizedTopicForCrossSiteMessaging
 	};
 
+	function isAuthorizedTopicForCrossSiteMessaging() {
```
guillegr123, line 88: It would be better to receive the topic name as a parameter, and check if it is included in the array of topics within this function, based on its name.

Also, do we want this check to be availlable for external parties?

- If so, it will be necessary to add the corresponding documentation file at [docs/monster-util](https://github.com/2600hz/monster-ui/tree/master/docs/monster-util)

- If not, then I would move this function to [core/app.js](https://github.com/2600hz/monster-ui/blob/master/src/apps/core/app.js)

docs/configuration.md
```diff
@@ -175,6 +175,13 @@ Key | Description | Type | Default | Required
 `brands.[].keyFunctions` | List key functions available per brand ([schema][combo-keys-schema]). | `Array` | | `false`
 `brands.[].lineKeys` | List of keys to be used as default line keys per brand. | `Array` | | `false`
 
+#### `crossSiteMessaging`
```
guillegr123, line 178: Please add an entry to the [Settings table above](https://github.com/2600hz/monster-ui/blob/a7eee0c7f85cd6fedce29366cddbc96c871e0d30/docs/configuration.md#settings), to indicate where the `crossSiteMessaging` should be included.

guillegr123 (changes requested):

guillegr123:
@ramandeepromana if possible, do a merge or rebase of this branch with `master`, in order to fix the CI builds.

src/apps/core/app.js
```diff
@@ -89,13 +90,42 @@ define(function(require) {
...
 				_.bind(self.loadIncludes, self)
 			], _.bind(self.loadAuth, self));
 		},
 
+		bindCrossSiteMessagingHandler: function(crossSiteMessaging) {
+			const { origin, topics } = crossSiteMessaging;
+
+			var handleCrossSiteMessages = function handleCrossSiteMessages(event) {
+				var activeApp = monster.apps.getActiveApp(),
+					triggerApp = event.data.split('.')[0];
```
guillegr123, line 109: Check if `event.data` is a string, before the split, to avoid errors caused by messages with a different payload. This can be checked by using `lodash#isString`.

src/apps/core/app.js
```diff
@@ -89,13 +90,42 @@ define(function(require) {
 			self.displayFavicon();
 			self.loadSVG();
 
+			if (crossSiteMessaging !== undefined) {
```
guillegr123, line 93: Use `lodash#isUndefined` to check if the value is undefined.

guillegr123 (changes requested):

docs/monster-util/isAuthorizedTopicForCrossSiteMessaging.md
```diff
@@ -0,0 +1,19 @@
+"title: isAuthorizedTopicForCrossSiteMessaging()
```
guillegr123, line 1: Rename the file to `isAuthorizedTopicForCrossSiteMessaging().md` to follow the naming convention, and add it to the file `docs/mkdocs/mkdocs.yml` under the `monster.util` entry, I think that may fix the Circle error.

ramandeepromana: Ah good eye/catch 👓

guillegr123 (approved):
Looks good to me!

PR #1044, opened 2023-01-16
MCAR-19: Expose feature set flags to hide Devices configuration in Call Recoding

ramandeepromana:


src/js/lib/monster.js
```diff
@@ -157,6 +157,16 @@ define(function(require) {
 						eventRouting: true
 					}
 				}
+			},
+			callRecording: {
+				28: {
+					configuration: {
+						manage: true,
```
guillegr123, line 164: Is this value (`callRecording.configuration.manage`) going to be read in any app? I don't see it read at [MCAR-19](https://github.com/2600hz/monster-ui-call-recording/pull/30/files)

ramandeepromana: Oh no and I added that as Devices section falls under Configuration 🤔 

My thinking was something along the lines of `callRecording: {` similar to `smartpbx: {` with `28: {` under both of them

& `configuration: {` section under `callRecording` with  option to `manage` as we will be working on that, followed by `devices: { manage:` as the falls under Configuration

guillegr123: Oh, OK, got it 👍

guillegr123 (approved):
Looks good to me!

PR #1045, opened 2023-01-16
MCAR-18: Expose feature set flags to hide Storage Setting in Call Recording

Isaac2600Hz:


guillegr123 (approved):
Lgtm!

guillegr123 (approved):
Lgtm!

ramandeepromana:
@guillegr123 @Isaac2600Hz shouldn't `storageSettings: { manage: }` be `false` instead of `true` for settings to hidden in the Call Recording application 🙂

guillegr123:
You are absolutely right @ramandeepromana, the storage settings will be available when the value is `true`, and hidden when the value is `false`. Is it the intended behavior to show the storage setting for the feature set number 28, @Isaac2600Hz?

PR #1047, opened 2023-01-18
MSPB-319: Added new setting/ability for disabling walkthrough

ramandeepromana:
Added new configuration setting/logic for disabling new user walkthrough.

### DIFF
--- a/src/apps/myaccount/app.js
+++ b/src/apps/myaccount/app.js
@@ -60,7 +60,8 @@ define(function(require) {
 					'name',
 					'full'
 				]
-			}
+			},
+			disableShowfirstUseWalkthrough: monster.config.whitelabel.disableShowfirstUseWalkthrough
 		},
 
 		mainContainer: '#myaccount',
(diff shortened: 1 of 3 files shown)


guillegr123 (changes requested):
It looks good! Just to avoid any possible confusion by having a long name, could you shorten them a bit from `disableShowfirstUseWalkthrough` to `disableFirstUseWalkthrough`?

src/apps/myaccount/app.js
```diff
@@ -270,7 +271,7 @@ define(function(require) {
 				if (currentAccount.hasOwnProperty('trial_time_left') && monster.config.api.hasOwnProperty('screwdriver')) {
 					monster.pub('auth.showTrialInfo', currentAccount.trial_time_left);
 				// Only direct them to my account walkthrough if it's not a trial account
-				} else if (self.hasToShowWalkthrough()) {
+				} else if (self.hasToShowWalkthrough() && !self.appFlags.disableFirstUseWalkthrough) {
```
guillegr123, line 274: Sorry, I just noticed the `hasToShowWalkthrough` function. It would be better to check the flag `self.appFlags.disableFirstUseWalkthrough` inside this function, because it is subscribed to the `myaccount.hasToShowWalkthrough` which means that the condition may be checked also outside of the `myaccount` app.

guillegr123 (approved):
Looks good to me!

PR #1048, opened 2023-01-25
MNUM-7: Add "GO" under "Caller ID Numbers" tab

Isaac2600Hz:


src/apps/common/submodules/numbers/numbers.js
```diff
@@ -811,6 +825,47 @@ define(function(require) {
 				});
 			};
 
+			var searchExternalNumber = function(searchString, parent) {
```
guillegr123, line 828: As the sub-accounts' numbers will no longer be pre-loaded, this needs to be refactored. Check how the function `searchListNumbers`  that is a few lines above works and use it as an example, or if possible use that same function, as it should work in a similar way for the 3 number tabs.

src/apps/common/submodules/numbers/numbers.js
```diff
@@ -1539,6 +1612,21 @@ define(function(require) {
 			});
 		},
 
+		listAccountExternalNumbers: function(accountId, success) {
```
guillegr123, line 1615: All the functions in a submodule should have the submodule name as prefix. In this case it would be something like `numbersListExternalNumbers`. Also, having the API call for `externalNumbers.list` as a separate function, it would be a good time to use it in here: https://github.com/2600hz/monster-ui/blob/master/src/apps/common/submodules/numbers/numbers.js#L1729

guillegr123 (changes requested):

PR #1052, opened 2023-03-01
[5.0] MACF-76: Fix Time Of Day Increment When Saving

Ramos95:
Creating new function validateEndOrBegeningOfGregorianDay to validate if date should be saved at the end of begining of gregorian date to avoid the issue where the saved date could apper as a fay before or after the one set on Time of Day edit on CallFlows due tu timezone hour offsets.

Related CallFlows PR: https://github.com/2600hz/monster-ui-callflows/pull/238


guillegr123:
This validation will be no longer necessary, as the issue was fixed by using the [`monster.util#getCurrentTimeZone()`](https://docs.2600hz.com/ui/docs/monster-util/getCurrentTimeZone%28%29/) helper to get the proper time zone before saving the date.

PR #1053, opened 2023-03-07
MSTR-240: Update auth account data due to the clients response changes

pcandia:


pcandia:
The accounts endpoint is now returning the data format in a different way, you can see the old way vs the new way.  We check for the param `superduper_admin` in `monster.util.isSuperDuper()`.  This current solution is working and requires minimal changes in the code, let me know what you guys think.

oldWay

newWay

guillegr123 (approved):
Lgtm!

PR #1055, opened 2023-03-07
MSPB-325: Fix URL encoding issue for phone number list query

ramandeepromana:
Update phone number list API URL to use proper URL encoding method for encoding brackets "[]"

### DIFF
--- a/src/js/lib/jquery.kazoosdk.js
+++ b/src/js/lib/jquery.kazoosdk.js
@@ -316,7 +316,7 @@
 				'delete': { verb: 'DELETE', url: 'accounts/{accountId}/phone_numbers/{phoneNumber}' },
 				'deleteBlock': { verb: 'DELETE', url: 'accounts/{accountId}/phone_numbers/collection' },
 				'identify': { verb: 'GET', url: 'accounts/{accountId}/phone_numbers/{phoneNumber}/identify' },
-				'list': { verb: 'GET', url: 'accounts/{accountId}/phone_numbers?filter_state=["in_service","port_in"]' },
+				'list': { verb: 'GET', url: 'accounts/{accountId}/phone_numbers?filter_state=' + encodeURIComponent('["in_service","port_in"]') },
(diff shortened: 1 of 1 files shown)


guillegr123 (changes requested):
I would suggest to use `encodeURI()` to clean the whole url instead, at [Line 613](https://github.com/2600hz/monster-ui/blob/a042c2013719a22037637ea8b8cae03fae79325b/src/js/lib/jquery.kazoosdk.js#L613), in order to avoid the same issue with any URL, like so:

```js
/* ... */
composedUrl = (methodSettings.apiRoot || settings.apiRoot) + methodInfo.url,
requestSettings = {
	url: encodeURI(composedUrl),
	verb: methodInfo.verb,
	data: {}
},
/* ... */
```

ramandeepromana:
@guillegr123 using `encodeURI()` over `encodeURIComponent(`) to clean the whole URL is definitely a better solution to avoid any similar issues with other/new URLs 🎈

ramandeepromana:
@guillegr123 however, upon encoding the complete URL for the requests I'm seeing weird behaviour/video attached below as well as not sure if other API/section(s) would be affected or not 😶‍🌫️ 

https://user-images.githubusercontent.com/78377672/223599578-de11f325-c13d-467a-8ce1-9dc933799e74.mov

guillegr123:
@ramandeepromana you are right, it seems that at this point the parameters embedded in the URL have not been replaced yet, so the curly braces also get encoded. It seems that it would be more complicated to encode the whole URL, because if it is done after all the parameters are replaced, they may be double-encoded, as they are already encoded at that point. Let's get back to your first solution using `encodeURIComponent` only for the specific filter parameter.

Sorry to have you running in circles 😅

guillegr123 (approved):
Lgtm!

ramandeepromana:
@guillegr123 oh no worries at all and thank you for the approval/thank you @pcandia as well 😄

PR #1056, opened 2023-03-07
[5.2] MSPB-325: Fix URL encoding issue for phone number list query

ramandeepromana:
Update phone number list API URL to use proper URL encoding method for encoding brackets "[]"


guillegr123 (approved):

pcandia (approved):
i verified that its working as expected.

PR #1074, opened 2023-04-04
[5.2] MSTR-232: Enable Marketplace Connector: Step 2 instructions needs to be corrected

claytonfinney:
Also fixed a small issue where {{ i18n.marketplace.linkage.stepSingin }} was in there twice


pcandia (approved):
lgtm

PR #1094, opened 2023-07-11
MSTR-263: Do not override the subscription selection when changing the auto recharge

pcandia:
…uto recharge

### DIFF
--- a/src/apps/myaccount/submodules/balance/balance.js
+++ b/src/apps/myaccount/submodules/balance/balance.js
@@ -286,14 +286,15 @@ define(function(require) {
 				}
 			}
 
-			if (_.get(data, 'account.topup.amount')) {
-				topupData.enabled = true;
-			}
-
 			if (data.account.hasOwnProperty('topup')) {
 				$.extend(true, topupData, data.account.topup);
 			}
 
+			if (_.get(data, 'account.topup.amount')) {
+				topupData.enabled = true;
+			}
+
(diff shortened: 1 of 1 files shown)


guillegr123 (changes requested):
The issue persists for this scenario:

1. Go to Control Center > Add credit, and make sure that auto recharge is turned off, and that one of the first 2 subscriptions is selected.
2. Reload the page (F5).
3. Go again to Control Center > Add credit, and enable auto recharge with any values.
4. Save, and check the request body for the POST used to update the account. You will see that `topup.monthly` has been removed.

i.e.: The issue still happens if it is the first time you go to Control Center > Add credit after the page has been loaded.

guillegr123 (changes requested):
There are still a couple of issues:

- The property `enabled` should not be included in the `topup` object.

image

- If you enable auto-recharge with some values, then disable it, and then enable it again but with different values, the new values are not saved. Instead, the previous values are used.

guillegr123 (changes requested):
Almost there! There is just one use case missing. If you try to change the existing values for auto-recharge, when you press the "Confirm Auto Recharge" button, the old values are saved, instead of the new values.

[Topup values not updated.webm](https://github.com/2600hz/monster-ui/assets/893877/0d50a34a-3679-4307-8239-dfa08ad9093b)

guillegr123 (approved):
Lgtm!

PR #1114, opened 2023-10-06
[5.3] MSTR-269: adding fields to display the cluster name and show placeholder text if name is not set

Ramos95:
Updating appstore views to include new fields:

- to display cluster name
- to display toggle to enable all in one cluster option
- to display cluster kazoo version

updating appstore request payload to use the same structure as the `marketplace` api.

image

image


pcandia (approved):
great job, lgtm

PR #1119, opened 2023-11-07
MSTR-267: Handle unsupported account default language

masmerino13:


src/js/lib/monster.apps.js
```diff
@@ -567,33 +567,19 @@ define(function() {
...
-					 */
 					if (!_.has(app.i18n, language)) {
-						console.info(language + ' isn\'t a supported language by this application: ' + app.name + ' switching automatically to ' + monster.defaultLanguage);
-						self.loadLocale(
-							app,
-							monster.defaultLanguage,
-							// Prepend null as we don't care if it errors out, the app can still load
-							_.partial(callback, null)
-						);
-
```
guillegr123, line 588: I think it is not necessary to pass `{ languageSupported: false }`

Se the other comment for details.

src/js/lib/monster.apps.js
```diff
@@ -624,9 +610,10 @@ define(function() {
 							language = _.find([
 								monster.config.whitelabel.language,
 								monster.defaultLanguage
-							], _.partial(_.includes, loadedLanguages));
+							], _.partial(_.includes, loadedLanguages)),
```
guillegr123, line 613: The problem with this condition is that the variable `loadedLanguages` is set after `app.data.i18n` is merged with the core app's i18n object above in [line 603](https://github.com/2600hz/monster-ui/pull/1119/files#diff-ba22e5d634130f5cee1042a194d3b9d7744331d3df283eae651061c58d326242R603)

I think it would be better to set `loadedLanguages` at [line 601](https://github.com/2600hz/monster-ui/pull/1119/files#diff-ba22e5d634130f5cee1042a194d3b9d7744331d3df283eae651061c58d326242R601) so it only includes the languages of the target app, so this way the parameter `languageSupported` won't be needed, and the code will work as originally intended.

guillegr123 (changes requested):

guillegr123 (approved):
Looks good to me!

PR #1123, opened 2023-11-15
MSTR-273: Preserve leading zeros in extension numbers

ramandeepromana:
Revised data processing to use the unaltered `number` variable instead of `parsedNumber`, as `parseInt()` automatically removes any leading zeros, and altered the HTML table column data type from `number` to `text` to counteract HTML's numerical data design which also strips leading zeros, thereby ensuring the preservation of leading zeros in extension numbers


pcandia (approved):
lgtm

PR #1129, opened 2023-12-08
MSTR-270:  adding new view to show when cluster was unlinked from portal and then return error when entering on monster ui

Ramos95:
handling when  monsterui marketplace connector returns  `401` error from `appex-server` when we open connector and showing  a view to complete the unlinking process and then return to the link `cluster` view.

Captura de pantalla 2023-12-21 101808

### DIFF
--- a/src/js/lib/monster.apps.js
+++ b/src/js/lib/monster.apps.js
@@ -636,17 +636,26 @@ define(function() {
 					_.partial(_.map, _, monster.normalizeUrlPathEnding),
 					_.partial(_.find, _, _.isString)
 				),
+				removeTrailingSlash = function(url) {
+					// Remove trailing '/'
+					return _.endsWith(url, '/') ? url.slice(0, -1) : url;
+				},
 				metadata = monster.util.getAppStoreMetadata(name),
-				externalUrl = getValidUrl([
+				externalUrl = _.flow(
+					getValidUrl,
+					removeTrailingSlash
+				)([
 					_.get(metadata, 'source_url'),
 					_.get(options, 'sourceUrl')
 				]),
 				hasExternalUrlConfigured = !_.isUndefined(externalUrl),
 				pathConfig = hasExternalUrlConfigured ? {
 					directory: externalUrl,
-					module: 'app-' + name
+					moduleRoot: 'apps/' + name,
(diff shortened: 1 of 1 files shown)


src/apps/appstore/i18n/en-US.json
```diff
@@ -88,6 +88,10 @@
 		},
 		"dangerDisableDialog": {
 			"content": "This will disable the Marketplace connector. The next time your KAZOO nodes start, apps purchased through the Marketplace will not be started. Do not forget to also remove those applications from your Kazoo nodes start up list."
+		},
+		"unlinkedCluster": {
+			"mainText": "Hey, if you unliked your cluster using the marketplace portal",
```
icehess, line 93: Change this to a better help text, ask Emily to help. Something along:

```
Hey, we unable to get your cluster information from Marketplace. It seems you unlinked your cluster using Marketplace Portal, if that is the case please click the button below to clear linkage status from cluster. You may want to re-link your cluster after.

If that is not the case check your Cluster ID is not changed and Cluster Secret is set and correct them.

If you think you Cluster ID or Secret is compromise please check the Marketplace place immediately and contact us for help
```

Run that with Emily first please.

src/apps/appstore/i18n/en-US.json
```diff
@@ -88,6 +88,10 @@
...
+			"buttonText":"Click here to cleanup cluster link state in your cluster"
```
icehess, line 94: Probably better to have a shorter/better label for this, how about`Complete Unlinking`?

icehess (changes requested):
If you can share a screenshot, this would be helpful. Otherwise just some comments regarding the messaging.

src/apps/appstore/i18n/en-US.json
Ramos95: updated with emily's text

src/apps/appstore/i18n/en-US.json
Ramos95: done 👍

icehess (approved):

guillegr123 (approved):

pcandia (approved):
lgtm

PR #1130, opened 2023-12-08
[5.3] MSTR-270: adding new view to show when cluster was unlinked from portal and then return error when entering on monster ui

Ramos95:
handling when monsterui marketplace connector returns 401 error from appex-server when we open connector and showing a view to complete the unlinking process and then return to the link cluster view.

Captura de pantalla 2023-12-21 101808


guillegr123 (approved):

pcandia (approved):
lgtm

PR #1146, opened 2024-03-20
[5.3] MSTR-286: Fix issue where expired credit cards are set to default

pcandia:


guillegr123 (approved):
Looks good! Due to time contsraints, only if possible, include the block for `requestPaymentMethod` in a `monster.waterfall` sequence, in order to avoid nesting too many callbacks.

guillegr123 (approved):

PR #1203, opened 2024-06-06
MSPB-365: Add system config endpoint

masmerino13:


src/js/lib/jquery.kazoosdk.js
```diff
@@ -588,6 +588,9 @@
 				'update': { 'verb': 'POST', 'url': 'accounts/{accountId}/match_lists/{matchListId}' },
 				'patch': { 'verb': 'PATCH', 'url': 'accounts/{accountId}/match_lists/{matchListId}' },
 				'delete': { 'verb': 'DELETE', 'url': 'accounts/{accountId}/match_lists/{matchListId}' }
+			},
+			configs: {
+				'get': { 'verb': 'GET', 'url': 'accounts/{accountId}/configs/kazoo_endpoint'}
```
guillegr123, line 593: I think that kazoo_endpoint should be a parameter, because I guess that this endpoint may also be used for other types of configs.

guillegr123 (changes requested):

guillegr123 (approved):

PR #1251, opened 2024-11-27
[5.3] MSTR-327: Fix Payment Method Pop-up After Successful ACH Setup

ramandeepromana:
- Added `hasVerifiedACH` check to prevent the payment method pop-up after successful ACH setup.
- Ensured the pop-up triggers only when necessary conditions are unmet.


guillegr123 (approved):
Lgtm!

PR #1268, opened 2025-03-10
MSTR-331: Add 'id' to the payload when creating a sip device

pcandia:


plau (dismissed):
Looks good

src/apps/common/submodules/chooseModel/chooseModel.js
```diff
@@ -260,7 +261,10 @@ define(function(require) {
 
 			templateDevice.find('.action-device').on('click', function() {
 				if (monster.ui.valid(templateDevice.find('#device_form'))) {
-					var formData = monster.ui.getFormData('device_form'),
+					var brandSelected = _.find(dataTemplate.brands, { id: selectedBrand }),
+						familySelected = _.find(_.get(brandSelected, 'families'), { id: selectedFamily }),
+						modelSelected = _.find(familySelected.models, { id: selectedModel }),
```
guillegr123, line 266: I would suggest to standardize these searches to something like this:
```suggestion
					var brandSelected = _.chain(dataTemplate).get('brands', []).find({ id: selectedBrand }).value(),
						familySelected = _.chain(brandSelected).get('families', []).find({ id: selectedFamily }).value(),
						modelSelected = _.chain(familySelected).get('models', []).find({ id: selectedModel }).value(),
```

guillegr123 (changes requested):

guillegr123 (approved):

PR #1272, opened 2025-05-21
MSTR-333: Hide cancel/close buttons when password is expired

pcandia:


src/apps/auth/app.js
```diff
@@ -396,8 +399,10 @@ define(function(require) {
 				if (err) {
 					monster.util.logoutAndReload();
 				} else {
+					var isPasswordExpired = _.get(results.user, 'is_password_expired', false);
+
```
guillegr123, line 403: Remove `is_password_expired` from user here, because it will not be needed anywhere else. This way, you will not need to remove it later in the other places.
```suggestion
delete results.user.is_password_expired;
```

guillegr123 (changes requested):

guillegr123 (approved):
Looks good to me!

PR #1399, opened 2026-07-17
[5.4] MSPB-413: Display device name if image doesn't exist when creating a SIP Phone

pcandia:


guillegr123 (dismissed):
Looks good to me!

